PR #13, opened 2018-07-14
Use markdown parser to handle code blocks properly

### LINKED ISSUE #7: Code blocks still broken in some cases
In this case, *test* is not converted, since the plugin thinks it's inside a fenced codeblock - but it's not, since the fences are in separate code blocks started by the four-space indentation:

```
    ```

test

    ```
```

---

Another case of thinking *test* is inside a fenced code block even though it isn't:

```
` ``` `
test
` ``` `
```

---

In this case, *test* is converted even though it is inside a code block (created by the four-space indentation:

```
    test
```

---

I think the cleanest solution here would be using a markdown parser, i.e. convert to an AST, apply the linkification on text that is neither part of an existing link, nor inside a code block nor inline code. Like this it wouldn't matter to the plugin how the code block is created at all.

### LINKED ISSUE #11: Double-linking may happen
With this rule:

    Pattern:  "(Mattermost)",
    Template: "[Mattermost](https://mattermost.com)",

this text:

    [  Mattermost  ](https://mattermost.com)

is converted to

    [  [Mattermost](https://mattermost.com)  ](https://mattermost.com)

---

This would also be solved by using the markdown parser (https://github.com/mattermost/mattermost-plugin-autolink/issues/7#issuecomment-400693623), when adding a check like this to the Inspect callback:

```go
switch node.(type) {
case *markdown.InlineLink:
	return false
}
```

(same needed for inline images and the reference versions)

ThiefMaster:
This fixes #7 and fixes #11.

Important: I did not do the `dep update` here since it would add noise to the PR and I don't know whether you usually include this in a PR or rather do it manually before merging a PR requiring more recent server code. It is nonetheless required for my changes to work as the currently vendorized server code does not contain my change from mattermost/mattermost-server#9067

Once mattermost/mattermost-server#9112 has been merged, the `foobar` in the tests should be replaced with `foo!bar` to ensure everything it works for a string that is currently split over multiple text nodes.

### DIFF
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -1,13 +1,14 @@
 package main
 
 import (
-	"strings"
+	"fmt"
 	"sync/atomic"
 
 	"github.com/mattermost/mattermost-server/mlog"
 
 	"github.com/mattermost/mattermost-server/model"
 	"github.com/mattermost/mattermost-server/plugin"
+	"github.com/mattermost/mattermost-server/utils/markdown"
 )
 
 // Plugin the main struct for everything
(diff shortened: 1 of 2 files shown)


ThiefMaster:
Thanks! Don't forget to update the vendorized server to the latest `master` - with my changes the plugin won't compile against the older server version that's currently bundled.

coreyhulen:
Thanks for the heads up.  Awesome on this @ThiefMaster!

PR #23, opened 2018-09-06
Fix regex for match following a comma, fixes #17

### LINKED ISSUE #17: Problem with following punctuation marks
I just tried one of the example:
`{
    "Pattern": "(EXA)(-)(?P<jira_id>\\d+)",
    "Template": "[EXA-${jira_id}](https://tracker.example.com/browse/EXA-${jira_id})"
}`

If I type `Check this EXA-12456 it's great` it correctly replaces it. But if I type `Check this EXA-12345, it's great` it doesn't. I don't see anything wrong with the pattern.

Is there a way to make this work or is it a bug which should be fixed?

lumbric:


lumbric:
This is a very experimental completely untested change, see #17 for further discussion.

PR #25, opened 2018-09-07
 Fix regex to allow brackets, fixes #24

### LINKED ISSUE #24: Problem with opening and closing brackets
This is related to #17 which fixes a bug where text was not autolinked if there was a following punctuation mark.

The same bug occurs if brackets are used. Like in `there is a bug (see EXA-1234)`.

lumbric:


ThiefMaster:
Shouldn't opening brackets at the beginning be allowed as well? For example, `check my last ticket (XXX-1234)` still doesn't autolink

lumbric:
> Shouldn't opening brackets at the beginning be allowed as well? For example, `check my last ticket (XXX-1234)` still doesn't autolink

@ThiefMaster yes, you are right see [here](https://github.com/mattermost/mattermost-plugin-autolink/issues/24#issuecomment-419568626) in #24.

Issue #26, opened 2018-09-20
Patterns with hyperlink can't be autolink


shawn111:
Hi,

I like this plugin, and just found this issue.

In the test cases and examples, we have below patterns.
```
 {
   "Pattern": "(https://mattermost.atlassian.net/browse/)(MM)(-)(?P<jira_id>\\d+)",
   "Template": "[MM-$jira_id](https://mattermost.atlassian.net/browse/MM-$jira_id)"
 },
```
It can pass the unittest.

However, it isn't working in the real services.
I guess the hyperlink is already be modified that cause the pattern not work.

CrowdfoxMKL:
I have to second this - I tried the samples from the README, but to no avail. I even tested this further and found out, as soon as the pattern is to be compared to a actual link in Mattermost, it doesn't get modified. 
As a sample try these to pattern/template combos:
```
                {
                    "Pattern": "(https:test)",
                    "Template": "found"
                },
                {
                    "Pattern": "(https://test)",
                    "Template": "found"
                },
```
As you will see, the first pattern matches if you type the pattern im Mattermost, the second pattern does not match.

I'd really like to use this, so I can clean up the conversations of our developers a bit :)

grubbins:
I also encountered this problem. It will affect anything that gets replaced with an `AutoLink` during markdown processing, which currently applies to the following URL schemes:
```golang
DefaultUrlSchemes = []string{"http", "https", "ftp", "mailto", "tel"}
```

The content of these is skipped by the plugin because it is no longer raw text. I believe I have a fix (PR on the way).

grubbins:
The issue can also be demonstrated by adding suitable cases to `plugin_test.go` which applies the markdown parser (unlike `autolinker_test.go`).

jespino:
Fixed in the PR #38

Issue #36, opened 2019-01-15
Updated posts are not processed


grubbins:
If a post is edited to fix a typo, replacements are not applied. The plugin should handle the `MessageWillBeUpdated` hook too.

grubbins:
I tried to fix this by adding that hook, but it is still not working :/
Not sure what the problem is at the moment.

hanzei:
Thanks for the report @grubbins.

I'm not sure that this is something we really want. The main problem is that users have no possibility to fix posts that where wrongly edited by the plugin.

grubbins:
Actually I think there are ways to fix posts. For example, if my text is replaced with an unwanted link, I can modify the text so that it does not match any more, or put it in a code block, or similar.

Conversely, consider using this plugin for replacing Jira ticket numbers with links (as I have set up). It is very confusing and irritating to users that this works on first posting, but does not work if they edit their post to add a Jira ticket number.

In our MM installation we will definitely want to keep the replacements on post updates. It could be a config option for each pattern?

hanzei:
> Actually I think there are ways to fix posts. For example, if my text is replaced with an unwanted link, I can modify the text so that it does not match any more, or put it in a code block, or similar.

But how does a user know, what the pattern is? The normal users doesn't even know this plugins exists and might not even not what a plugin is. The changes to a post happen "magically" and users don't know why. Adding a config option doesn't fix this problem.

@coreyhulen What do you think?

jespino:
Maybe we can add a "pattern" flag to allow to work on updates. Normally a lot of patterns are idempotent, so is not a big deal, but for that cases where you are not able to handle properly editions, you can disable it per-pattern (or maybe enable it if we set it as false by default).

hanzei:
I would rather add a undo function. User should have the control over there own posts.

My fear is that posts get mangled and a user doesn't know how to fix this, like in #19.

hanzei:
I see that an undo function is out of scope.

But flagging pattern can still result in the problem I described above. Are the other solutions you can think of? I'm running out of ideas.

jespino:
I'll go for the flagging pattern approach for now, which will allow our uses to enable it for editions under his own responsability. Is better than nothing, and probably we can add another strategy later.

levb:
To be clear on the proposed course of action:
1. Add a `EnableOnUpdate` default `false` to Link config properties.
[...]

PR #37, opened 2019-01-15
Support autolinks and post updates

grubbins:
I believe this will fix issues #26 and #36.

There is one TODO in there which is not critical: if the raw text contains www.somewhere.com, an autolink is created with destination http://www.somewhere.com, and the plugin logs an error because the destination text and original text don't match (and no replacements are applied). This behaviour is acceptable I think but better to suppress the warning.


jespino:
I'm closing this in favor of #41, which is a commit based on your changes, but only with the post update part. Actually, I set you as the "Author" of the commit.

PR #38, opened 2019-01-26
Handling autolink nodes too, not only text nodes

jespino:


jespino:
Sure

jespino:
Actually, is not there, must be in the `plugin_test.go`, because `autolink.go` only deals with the replacement of text, don't care about markdown or nodes.

jespino:
Done :)

jespino:
I added the `t.Run` in the text to be more "verbose" about which cases are working and which are failing.

server/plugin.go
```diff
@@ -60,14 +60,28 @@ func (p *Plugin) MessageWillBePosted(c *plugin.Context, post *model.Post) (*mode
...
-		if textNode, ok := node.(*markdown.Text); ok {
-			startPos, endPos := textNode.Range.Position+offset, textNode.Range.End+offset
-			origText := postText[startPos:endPos]
+		origText := ""
+		startPos := 0
+		endPos := 0
+
+		if autolinkNode, ok := node.(*markdown.Autolink); ok {
+			startPos, endPos = autolinkNode.RawDestination.Position+offset, autolinkNode.RawDestination.End+offset
+
```
ThiefMaster: i'd remove this linebreak or add it in the `else if` block as well for the sake of consistency

ThiefMaster:
> Actually, is not there, must be in the plugin_test.go, because autolink.go only deals with the replacement of text, don't care about markdown or nodes.

What about merging the two sets of test cases, moving them to a single place, and then using same set of test data for both `plugin_test` and `autolink_test`?

jespino:
@ThiefMaster I think it can be done, but I'm not 100% sure about doing it, I like to have the test cases inside the test itself, if I have to import shared testcases elsewhere will look weird. Anyway, if you have an strong opinion here, you can make a PR and we can discuss it seeing how it ends.

PR #39, opened 2019-01-29
Fix replacing autolinked text

ThiefMaster:
This was most likely broken by the change in mattermost/mattermost-server#9151

### DIFF
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -48,7 +48,9 @@ func (p *Plugin) MessageWillBePosted(c *plugin.Context, post *model.Post) (*mode
 	postText := post.Message
 	offset := 0
 	markdown.Inspect(post.Message, func(node interface{}) bool {
-		switch node.(type) {
+		var textNode *markdown.Text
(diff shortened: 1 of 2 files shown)


server/plugin.go
```diff
@@ -58,9 +60,13 @@ func (p *Plugin) MessageWillBePosted(c *plugin.Context, post *model.Post) (*mode
 			return false
 		case *markdown.ReferenceImage:
 			return false
+		case *markdown.Autolink:
+			textNode = typedNode.Children[0].(*markdown.Text)
```
ThiefMaster: I don't particularly like this line - but the server uses an array for Autolink.Children even though those nodes always have exactly one Text child and nothing else.

ThiefMaster:
Oops, just checked the commits but not the open PRs!

jespino:
No problem, actually there was other PR solving the exact same thing before, but was fixing to things at a time, and I think is better to split them in two different PRs.

PR #41, opened 2019-01-31
Allowing autolink replacement on update too

jespino:
related to issue #36

### DIFF
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -41,9 +41,7 @@ func (p *Plugin) OnConfigurationChange() error {
 	return nil
 }
 
-// MessageWillBePosted is invoked when a message is posted by a user before it is committed
-// to the database.
-func (p *Plugin) MessageWillBePosted(c *plugin.Context, post *model.Post) (*model.Post, string) {
+func (p *Plugin) processPost(c *plugin.Context, post *model.Post) (*model.Post, string) {
 	links := p.links.Load().([]*AutoLinker)
 	postText := post.Message
 	offset := 0
@@ -97,4 +95,16 @@ func (p *Plugin) MessageWillBePosted(c *plugin.Context, post *model.Post) (*mode
 	post.Message = postText
 
 	return post, ""
+}
+
+// MessageWillBePosted is invoked when a message is posted by a user before it is committed
+// to the database.
+func (p *Plugin) MessageWillBePosted(c *plugin.Context, post *model.Post) (*model.Post, string) {
+	return p.processPost(c, post)
+}
+
+// MessageWillBeUpdated is invoked when a message is updated by a user before it is committed
+// to the database.
+func (p *Plugin) MessageWillBeUpdated(c *plugin.Context, post *model.Post, _ *model.Post) (*model.Post, string) {
+	return p.processPost(c, post)
 }
--- a/server/plugin_test.go
+++ b/server/plugin_test.go
@@ -38,8 +38,17 @@ func TestPlugin(t *testing.T) {
 func TestSpecialCases(t *testing.T) {
 	links := make([]*Link, 0)
 	links = append(links, &Link{
+		Pattern:  "https://mattermost.com",
+		Template: "[the mattermost portal](https://mattermost.com)",
+	}, &Link{
 		Pattern:  "(Mattermost)",
 		Template: "[Mattermost](https://mattermost.com)",
+	}, &Link{
+		Pattern:  "https://mattermost.atlassian.net/browse/MM-(?P<jira_id>\\d+)",
+		Template: "[MM-$jira_id](https://mattermost.atlassian.net/browse/MM-$jira_id)",
+	}, &Link{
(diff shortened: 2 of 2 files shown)


server/plugin.go
```diff
@@ -98,3 +96,15 @@ func (p *Plugin) MessageWillBePosted(c *plugin.Context, post *model.Post) (*mode
...
+
+// MessageWillBePosted is invoked when a message is posted by a user before it is committed
+// to the database.
+func (p *Plugin) MessageWillBePosted(c *plugin.Context, post *model.Post) (*model.Post, string) {
+	return p.processPost(c, post)
+}
+
+// MessageWillBeUpdated is invoked when a message is updated by a user before it is committed
+// to the database.
+func (p *Plugin) MessageWillBeUpdated(c *plugin.Context, post *model.Post, _ *model.Post) (*model.Post, string) {
```
jespino: This can break some existing use cases, but I think it the expected behavior for the most cases.

server/plugin.go
```diff
@@ -98,3 +96,15 @@ func (p *Plugin) MessageWillBePosted(c *plugin.Context, post *model.Post) (*mode
...
 	return post, ""
 }
+
+// MessageWillBePosted is invoked when a message is posted by a user before it is committed
+// to the database.
+func (p *Plugin) MessageWillBePosted(c *plugin.Context, post *model.Post) (*model.Post, string) {
+	return p.processPost(c, post)
+}
+
+// MessageWillBeUpdated is invoked when a message is updated by a user before it is committed
```
lieut-data: Hmm, I'm worried about this change: the autolink plugin frequently breaks my edits, and my only course to recovery is to edit and fix it manually. With this change, it sounds like that path would be taken from me.

server/plugin_test.go
```diff
@@ -38,8 +38,17 @@ func TestPlugin(t *testing.T) {
 func TestSpecialCases(t *testing.T) {
 	links := make([]*Link, 0)
 	links = append(links, &Link{
+		Pattern:  "https://mattermost.com",
+		Template: "[the mattermost portal](https://mattermost.com)",
```
lieut-data: What is "the mattermost portal? Why do we need to auto-link this?

server/plugin_test.go
```diff
@@ -148,6 +157,15 @@ func TestSpecialCases(t *testing.T) {
 		}, {
 			"![  Mattermost  ][1]\n\n[1]: https://mattermost.com/example.png",
 			"![  Mattermost  ][1]\n\n[1]: https://mattermost.com/example.png",
+		}, {
+			"Why not visit https://mattermost.com?",
```
lieut-data: I know these are just unit tests, but what happens for:
```
Why not visit https://mattermost.com?important_query_string_parameter
```

server/plugin_test.go
```diff
@@ -38,8 +38,17 @@ func TestPlugin(t *testing.T) {
 func TestSpecialCases(t *testing.T) {
 	links := make([]*Link, 0)
 	links = append(links, &Link{
+		Pattern:  "https://mattermost.com",
+		Template: "[the mattermost portal](https://mattermost.com)",
+	}, &Link{
 		Pattern:  "(Mattermost)",
 		Template: "[Mattermost](https://mattermost.com)",
+	}, &Link{
+		Pattern:  "https://mattermost.atlassian.net/browse/MM-(?P<jira_id>\\d+)",
```
lieut-data: Is this materially different than line 56?

server/plugin.go
hanzei: Could we please keep the discussion _if_ we want this in https://github.com/mattermost/mattermost-plugin-autolink/issues/36? There have been some interesting points there.

jespino: I get your point. I'm evaluating the posibility of make "updateOnEdit" an option for the pattern. I mean, if you feel comfortable and safe with the pattern, and you know the patter is idempotent, you can enable the updateOnEdit. If not, you can disable it. But is really annoying to not have that possibility.

server/plugin_test.go
jespino: This is just a test example. Basically you can have this autolinked, but you can have this "auto-linked with a specific text" instead of the url.

server/plugin_test.go
jespino: Yes, there is no different, I think I fixed that in other PR.

server/plugin_test.go
jespino: This is a matter of your regular expression. In this case, this test simply do not capture any query param, but is up to you to write the proper regular expresion.

server/plugin_test.go
jespino: nevermind, I going to remove that test.

lieut-data:
@jespino / @hanzei, can we close this PR until the bigger discussion is addressed?

hanzei:
I'm fine with that.

jespino:
I going to close it and we can rethink it later.

PR #42, opened 2019-03-01
introducing circleci

cpanato:
Based on the dev meeting we are moving to circleci


scottleedavis:
Looking forward to this one!

cpanato:
@lieut-data will add you as a reviewer only in this PR if all good, will replicate to others

Makefile
```diff
@@ -49,7 +49,9 @@ endif
 govet:
 ifneq ($(HAS_SERVER),)
 	@echo Running govet
-	@$(GO) vet $$(go list ./server/...) || exit 1
+	$(GO) get golang.org/x/tools/go/analysis/passes/shadow/cmd/shadow
+	$(GO) vet $(GOFLAGS) $$(go list ./server/...) || exit 1
```
lieut-data: I realized now that the `exit 1` here is actually redundant -- can we strip it out? The `Makefile` will automatically terminate the target if any of the individual steps fail.

cpanato: remove for both? or just in the line 53? @lieut-data

cpanato: done @lieut-data

cpanato:
ping @hanzei will wait your review and then replicate to the others

Makefile
```diff
@@ -49,7 +49,9 @@ endif
 govet:
 ifneq ($(HAS_SERVER),)
 	@echo Running govet
-	@$(GO) vet $$(go list ./server/...) || exit 1
+	$(GO) get golang.org/x/tools/go/analysis/passes/shadow/cmd/shadow
+	$(GO) vet $(GOFLAGS) $$(go list ./server/...)
```
hanzei: I forgot that we don't use `$GOFLAGS` in this Makefile. :man_facepalming: 

But I still think `$GOFLAGS` can be useful.

@lieut-data @cpanato Thoughts on just adding adding a global variable `$GOFLAGS` at the start of the file and use it in everywhere?

hanzei:
@cpanato Please wait with the replicating until the above question is resolved.

Makefile
cpanato: sounds good to me

lieut-data: @hanzei, can you share a use case for `$(GOFLAGS)` in the context of `go vet`?

I note that on https://github.com/golang/go/issues/21711#issuecomment-326197693:
> Go vet never builds, so any try build-time flags are pointless...

Furthermore, on https://golang.org/cmd/go/#hdr-Report_likely_mistakes_in_packages:
> The build flags supported by go vet are those that control package resolution and execution, such as -n, -x, -v, -tags, and -toolexec. For more about these flags, see 'go help build'.

So would a generic `$(GOFLAGS)` actually be meaningful in this context? I'm wondering if we just remove it.

I'm also hesitant to tack on changes unrelated to the primary mission of building on Go 1.12 -- I think it's perfectly fine to make follow up, plugin-wide changes as necessary, but thinking we should probably cut this off and just fix the Go 1.12 issues now.

hanzei: Thanks for investigation into this @lieut-data! You answer is more in depth then it's expected for this small topic.

I was under the impression that this would would be more helpful mainly because the server also uses them. https://github.com/mattermost/mattermost-server/blob/master/Makefile#L294

But your right, this is out of scope for now. Let's defer this for later and just remove `$(GOFLAGS)` from the two line.

lieut-data: Sounds good, @hanzei! I admit I glossed right past `$(GOFLAGS)` in the server on the original review. There might be a case for removing them there as well. There's a ton of cruft that just fades into the background now for me ;P

cpanato: will do all the things, bear with me

Issue #10, opened 2018-06-27
Team/channel-scoped rules


ThiefMaster:
Especially when autolinking JIRA tickets (`FOO-\d+`) or GitHub issues (`#\d+`), chances are good that they only matter in certain channels/teams. and make no sense somewhere else (or should link to a different target),

Currently the only option is to add only rules that are useful instance-wide and not do any autolinking for anything that's less global. It would be nice if rules could be defined on the team and/or channel level.

outofcontrol:
This is highly desired in our Gitlab install also, where we have many groups with many repositories in each.

scottleedavis:
Hello @hanzei  I would like to contribute this.  My company uses this, and it would be helpful for us to only allow certain autolinks in certain channels.

hanzei:
Thanks @scottleedavis for working on this!

scottleedavis:
@ThiefMaster @outofcontrol Do either of you have any use cases for a channel scope, across teams?

https://github.com/mattermost/mattermost-plugin-autolink/pull/43#issuecomment-468860186

ThiefMaster:
Channel scope without mentioning a team: no, seems somewhat useless, unless someone wants e.g. off-topic to have the same rules in all teams. seems not very useful though.

`teamname/channelname` is very useful though, e.g. for a development team that wants to autolink `#123` or `project#123` to their github issue tracker without affecting all other channels. I would very much use that one!

scottleedavis:
Fantastic feedback, thank you @ThiefMaster

outofcontrol:
@scottleedavis Across teams no. Within teams how @ThiefMaster describes it, would be quite nice to have.

scottleedavis:
alright, majority wins.  :)

jasonblais:
@scottleedavis Does your PR (#43) cover this open issue?

scottleedavis:
yes @jasonblais

PR #43, opened 2019-03-02
Scope added to autolink filter

scottleedavis:
Adds Team and Channel Scope to plugin config, and restricts autolinks under given Scope if present.

Contributing for https://github.com/mattermost/mattermost-plugin-autolink/issues/10


hanzei:
What do you think of consolidating TeamScope and ChannelScope inside one scope and letting it have it be like `$TEAM/$CHANNEL`. Otherwise you can run into problems like if a channel is present in two teams.

scottleedavis:
That could work, though what if it was desired to scope by channel name and not team?

hanzei:
I feel this is a rare use case. But then you can just add a scope for every team.

README.md
```diff
@@ -28,11 +28,13 @@ Below is an example of regexp patterns used for autolinking, modified in the `co
             "links": [
                 {
                     "Pattern": "(LHS)",
-                    "Template": "[LHS](https://docs.mattermost.com/process/training.html#lhs)"
+                    "Template": "[LHS](https://docs.mattermost.com/process/training.html#lhs)",
+                    "Scope": ["test/off-topic"]
```
hanzei: Just to make the example clear:
```suggestion
                    "Scope": ["team1/off-topic"]
```

README.md
```diff
@@ -28,11 +28,13 @@ Below is an example of regexp patterns used for autolinking, modified in the `co
...
                 },
                 {
                     "Pattern": "(RHS)",
-                    "Template": "[RHS](https://docs.mattermost.com/process/training.html#rhs)"
+                    "Template": "[RHS](https://docs.mattermost.com/process/training.html#rhs)",
+                    "Scope": ["another-test/town-square"]
```
hanzei: Let's add an example for a team scope
```suggestion
                    "Scope": ["team2"]
```

server/plugin.go
```diff
@@ -82,10 +82,17 @@ func (p *Plugin) MessageWillBePosted(c *plugin.Context, post *model.Post) (*mode
 
 		if origText != "" {
 			newText := origText
+
+			channel, _ := p.API.GetChannel(post.ChannelId)
```
hanzei: 3/5 that you should not ignore the error here.

server/plugin.go
```diff
@@ -82,10 +82,17 @@ func (p *Plugin) MessageWillBePosted(c *plugin.Context, post *model.Post) (*mode
...
+			team, _ := p.API.GetTeam(channel.TeamId)
```
hanzei: Same here.

server/plugin.go
```diff
@@ -98,3 +105,16 @@ func (p *Plugin) MessageWillBePosted(c *plugin.Context, post *model.Post) (*mode
 
 	return post, ""
 }
+
+func contains(channel string, team string, list []string) bool {
```
hanzei: Mind moving this into a `util.go` file?

server/plugin.go
```diff
@@ -82,10 +82,17 @@ func (p *Plugin) MessageWillBePosted(c *plugin.Context, post *model.Post) (*mode
...
 			newText := origText
+
+			channel, _ := p.API.GetChannel(post.ChannelId)
+			team, _ := p.API.GetTeam(channel.TeamId)
+
 			for _, l := range links {
-				newText = l.Replace(newText)
+				if len(l.link.Scope) == 0 {
+					newText = l.Replace(newText)
+				} else if contains(channel.Name, team.Name, l.link.Scope) {
```
hanzei: It's kind of hard to read. That do you think about using 
```suggestion
				if len(l.link.Scope) != 0 && !contains(channel.Name, team.Name, l.link.Scope) {
					continue
				}
				newText = l.Replace(newText)
```

instead?

server/plugin.go
```diff
@@ -98,3 +105,16 @@ func (p *Plugin) MessageWillBePosted(c *plugin.Context, post *model.Post) (*mode
 
 	return post, ""
 }
+
+func contains(channel string, team string, list []string) bool {
+	for _, channelTeam := range list {
+		channelTeamSplit := strings.Split(channelTeam, "/")
+		if len(channelTeamSplit) == 2 {
```
hanzei: Maybe log an error if `len(channelTeamSplit) != 2`

server/plugin.go
```diff
@@ -98,3 +105,16 @@ func (p *Plugin) MessageWillBePosted(c *plugin.Context, post *model.Post) (*mode
...
+			if strings.EqualFold(channelTeamSplit[0], team) && strings.EqualFold(channelTeamSplit[1], channel) {
+				return true
+			}
+		}
+
```
hanzei: Nit: Remove this line

README.md
hanzei: Mind adding an example for multiple scope per rule?

server/plugin.go
scottleedavis: If there is no Scope provided in configuration, == 0 is a valid check.

Let me know if I missed something.

server/plugin.go
scottleedavis: logging for != 1 & != 2.  (1 being the team only example you provided earlier)

server/plugin.go
hanzei: Maybe we misunderstand each other.

I'm proposing to check first, if a scope is set and then if the scope is meet. The default case should be to apply the replacer.

server/plugin.go
```diff
@@ -83,8 +82,14 @@ func (p *Plugin) MessageWillBePosted(c *plugin.Context, post *model.Post) (*mode
 		if origText != "" {
 			newText := origText
 
-			channel, _ := p.API.GetChannel(post.ChannelId)
-			team, _ := p.API.GetTeam(channel.TeamId)
+			channel, cErr := p.API.GetChannel(post.ChannelId)
+			if cErr != nil {
+				return false
```
hanzei: Please log an error.

server/plugin.go
scottleedavis: Will you block on this?

hanzei: If you are at least 4/5, that you code is well the current way, no. I'm just 2/5.

scottleedavis: I feel that handling 

No Scope in configuration
or
Scope contains team & channel
or 
Scope contains team

I am 4/5 I would change it if you found a more elegant solution to the above logic.

hanzei: Could you please point out, what is wrong on the solution pointed out above?

hanzei: Sorry, but now I want ambitious to figure out "the right solution".

server/plugin_test.go
```diff
@@ -49,16 +60,35 @@ func TestSpecialCases(t *testing.T) {
 	}, &Link{
 		Pattern:  "(foo!bar)",
 		Template: "fb",
+	}, &Link{
+		Pattern: "(example)",
+		Template: "test",
```
scottleedavis: should change to team

server/plugin_test.go
```diff
@@ -49,16 +60,35 @@ func TestSpecialCases(t *testing.T) {
...
+		Scope: []string{"test/off-topic"},
+	}, &Link{
+		Pattern: "(example)",
+		Template: "test",
```
scottleedavis: also team

README.md
```diff
@@ -28,11 +28,13 @@ Below is an example of regexp patterns used for autolinking, modified in the `co
...
                     "Pattern": "(LHS)",
-                    "Template": "[LHS](https://docs.mattermost.com/process/training.html#lhs)"
+                    "Template": "[LHS](https://docs.mattermost.com/process/training.html#lhs)",
+                    "Scope": ["team1/off-topic"]
                 },
                 {
                     "Pattern": "(RHS)",
-                    "Template": "[RHS](https://docs.mattermost.com/process/training.html#rhs)"
+                    "Template": "[RHS](https://docs.mattermost.com/process/training.html#rhs)",
+                    "Scope": ["team2", "team3", "team4/town-square"]
```
scottleedavis: I think I put this here while thinking about what I wanted, without adding a note to implement it.

server/util.go
```diff
@@ -0,0 +1,26 @@
...
+	"strings"
+
+	"github.com/mattermost/mattermost-server/mlog"
+)
+
+func contains(team string, channel string, list []string) bool {
+	for _, channelTeam := range list {
+		channelTeamSplit := strings.Split(channelTeam, "/")
+		if len(channelTeamSplit) == 2 {
+			if strings.EqualFold(channelTeamSplit[0], team) && strings.EqualFold(channelTeamSplit[1], channel) {
```
scottleedavis: need to check for ',' string arrays.

README.md
scottleedavis: actually, i don't like it.  removing.

server/manifest.go
```diff
@@ -5,5 +5,5 @@ var manifest = struct {
 	Version string
 }{
 	Id:      "mattermost-autolink",
-	Version: "0.4.0",
+	Version: "0.5.0",
```
hanzei: Please revert this. I like to have things clean and use a seperate PR for this.

PR #48, opened 2019-04-04
Add examples for autolink plugin

jasonblais:
I'm aware that some of these are in the usage list above, but I want to start a list of examples that we share with communities and build it out over time.

@jespino

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,8 +93,8 @@ Below is an example of regexp patterns used for autolinking, modified in the `co
 ## Examples
 
 1. Autolinking `Ticket ####:text with alphanumberic characters and spaces` to a ticket link. Use:
-  - Pattern: `(?i)(ticket )(?P<ticket_id>)( : )(?P<ticket_info>.*)`
-  - Template: `(ticket ${ticket_id}:${ticket_info})`
+  - Pattern: `(?i)(ticket )(?P<ticket_id>.+)(:)(?P<ticket_info>.*)`, or if the ticket_id is a number, then `(?i)(ticket )(?P<ticket_id>\d+)(:)(?P<ticket_info>.*)`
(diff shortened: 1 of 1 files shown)


jespino:
@jasonblais I like this examples, but collides a little with the "usage example", I think is nicer to have a bunch of examples in this format that you have proposed, and try to make the "configuration example" shorter.

README.md
```diff
@@ -89,3 +89,17 @@ Below is an example of regexp patterns used for autolinking, modified in the `co
     }
 },
 ```
+
+## Examples
+
+1. Autolinking `Ticket ####:text with alphanumberic characters and spaces` to a ticket link. Use:
+  - Pattern: `(?i)(ticket )(?P<ticket_id>)( : )(?P<ticket_info>.*)`
```
jespino: To autolink that, you need to change it to `(?i)(ticket )(?P<ticket_id>.+)(:)(?P<ticket_info>.*)` or maybe if you are sure that the ticket_id is a number you can use `(?i)(ticket )(?P<ticket_id>\d+)(:)(?P<ticket_info>.*)`

README.md
```diff
@@ -89,3 +89,17 @@ Below is an example of regexp patterns used for autolinking, modified in the `co
...
+  - Template: `(ticket ${ticket_id}:${ticket_info})`
```
jespino: If you want to create a link, you have to point to a url, something like `[Ticket ${ticket_id}: ${ticket_info}](https://github.com/mattermost/mattermost-server/issues/${ticket_id})`

jasonblais:
@jespino Thanks! Updated.

> I think is nicer to have a bunch of examples in this format that you have proposed, and try to make the "configuration example" shorter.

Agree. The configuration example is good, but it's not obvious what the examples do. Hoping the examples section can be used instead long term.

PR #50, opened 2019-04-18
Fix crash on DMs

ThiefMaster:
This should ideally have tests (I don't have time for that atm) and it might be better to check the message type instead, but I needed a quick fix anyway so here's a PR with my fix.


hanzei:
Just to double check: This is a regression from v0.4.0, correct?

ThiefMaster:
yes, the bug got introduced in #43

jespino:
I'll try to publish a 0.5.1 release later today

jespino:
@ThiefMaster @hanzei Published the 0.5.1 version, thanks a lot!

PR #56, opened 2019-06-02
Bump release 1.0.0

levb:


plugin.json
```diff
@@ -1,8 +1,8 @@
 {
   "id": "mattermost-autolink",
   "name": "Autolink",
+  "version": "1.0.0",
   "description": "Automatically rewrite text matching a regular expression into a Markdown link.",
-  "version": "0.5.1",
```
hanzei: Very, very nit: Why did you changed the order? The previous order was in line with https://developers.mattermost.com/extend/plugins/manifest-reference/.

levb: I'll change it back, was an accident, the dog ate my homework

PR #71, opened 2019-08-02
/autolink command to edit configuration

levb:
This PR was prompted by a plethora of tickets for autolink config changes on the community servers. Since there is no UI in the foreseeable future, and modifying configuration otherwise is difficult, I figured these would be useful tools for the more advanced administrators.

The command can be turned off entirely with a config setting in the system console.

`/autolink help` output:

###### Mattermost Autolink Plugin Administration
<linkref> is either the Name of a link, or its number in the /autolink list output. A partial Name can be specified, but some commands require it to be uniquely resolved.

- `/autolink list` - list all configured links.
- `/autolink list <linkref>` - list a specific link.
- `/autolink test <linkref> test-text...` - test a link on a sample.
- `/autolink enable <linkref>` - enable a link.
- `/autolink disable <linkref>` - disable a link.
- `/autolink add <name>` - add a new link, named <name>.
- `/autolink delete <linkref>` - delete a link.
- `/autolink set <linkref> <field> value...` - sets a link's field to a value. The entire command line - after <field> is used for the value, unescaped, leading/trailing whitespace trimmed.

Example:

```
/autolink add Visa
/autolink disable Visa
/autolink set Visa Pattern (?P<VISA>(?P<part1>4\d{3})[ -]?(?P<part2>\d{4})[ -]?(?P<part3>\d{4})[ -]?(?P<LastFour>[0-9]{4}))
/autolink set Visa Template VISA XXXX-XXXX-XXXX-$LastFour
/autolink set Visa WordMatch true
/autolink test Visa 4356-7891-2345-1111 -- (4111222233334444)
/autolink enable Visa
```


server/config.go
```diff
@@ -1,15 +1,89 @@
...
+		if err != nil {
+			mlog.Error(fmt.Sprintf("Error creating autolinker: %+v: %v", c.Links[i], err))
+		}
+	}
+
+	p.updateConfig(func(conf *Config) {
+		*conf = c
+	})
+
+	go func() {
```
levb: The `go func() {}` here is a suggestion to work around https://github.com/mattermost/mattermost-server/pull/11780 which may not be available until mattermost-server 5.15 or later. Thanks @lieut-data for the suggestion.

server/link.go
```diff
@@ -0,0 +1,148 @@
...
+
+// Compile compiles the link's regular expression
+func (l *Link) Compile() error {
+	if l.Disabled || len(l.Pattern) == 0 || len(l.Template) == 0 {
+		return nil
+	}
+
+	pattern := l.Pattern
+	template := l.Template
+	multipass := false
```
levb: This (`multipass`) probably warrants a comment. Previously, word boundaries were defined as custom regexp's. These patterns consume characters, so they can not be used with `ReplaceAll` functions. Example, if the pattern is defined as `"X"`, `"foo X bar"` the X will match, but in `"foo X X bar"` only the first X matches, the match consumes the subsequent space, and then the second X doesn't match.

see the `Replace()` method for more.

server/command.go
```diff
@@ -0,0 +1,345 @@
...
+	case "false", "off":
+		return false, nil
+	}
+	return false, errors.Errorf("Not a bool, %q", arg)
+}
+
+func saveConfigLinks(p *Plugin, links []Link) error {
+	conf := p.getConfig()
+	conf.Links = links
+	appErr := p.API.SavePluginConfig(conf.ToConfig())
```
levb: `.ToConfig()` here converts the `Link` structs into a tree of `map[string]interface{}` and `[]interface{}` for GOB/RPC compatibility.

aaronrothschild:
I'm surprised there's no interactive regex editor.  ;)

Is there a spinmint or test server I can try this code out on interactively?

levb:
@aaronrothschild I spun one up on the related PR, https://github.com/mattermost/mattermost-server/pull/11780

aaronrothschild:
@levb some questions/comments:
- In `/autolink list` we should display if they are active or not
- in `/autolink` help results there is `/autolink test Vi 4356-7891-2345-1111 -- (4111222233334444)` the `Vi` should presumably be `Visa`
- What does "WordMatch" flag do?
- What is "field" in this command used for? "/autolink set <linkref> <field> value... - sets a link's field to a value. The entire command line after <field> is used for the value, unescaped, leading/trailing whitespace trimmed."

levb:
@aaronrothschild 

> * In `/autolink list` we should display if they are active or not

it kind of does, the disabled links are crossed out

> * in `/autolink` help results there is `/autolink test Vi 4356-7891-2345-1111 -- (4111222233334444)` the `Vi` should presumably be `Visa`

`Vi` is a partial match, should work.

> * What does "WordMatch" flag do?

Uses the more standard `\b` word boundaries. "\b - at ASCII word boundary (\w on one side and \W, \A, or \z on the other)". We historically implemented a custom set for the left and right boundaries, and it can (has) lead to confusion. I think we should eventually move towards making \b the default, and deprecating the old-style separators, but not yet.

> * What is "field" in this command used for? "/autolink set   value... - sets a link's field to a value. The entire command line after  is used for the value, unescaped, leading/trailing whitespace trimmed."

The name of the field. Like, for `Name`, `Pattern`, or `DisableNonWordPrefix`

DHaussermann:
@levb or @aaronrothschild  I see that this PR is still open. Can you please advise if we intend to package this for the 5.14 Release?

levb:
This is for _after_ 5.14 @DHaussermann, but I am merging this now, and will cut a release shortly. I'd like to deploy it onto community this week, once we confirm it on ci-extensions.

server/link.go
levb: With better naming, didn't need to comment as much

PR #75, opened 2019-09-04
Add `/autolink import/export` commands

### LINKED ISSUE #73: Add `/autolink import/export` capability
Some ideas for how it can be implemented:
- a browser popup/window that implements upload/download.
- export can create an uploaded file in Mattermost, import can read from a previously uploaded Mattermost file.

See #71 that introduced the config read/write capability.

adarj:
Enables importing and exporting config files using the following Autolink commands:

- `/autolink import` - imports the Autolink configuration file uploaded in the previous post
- `/autolink export` - exports the Autolink configuration as `autolink-config.json` to the current channel

Closes #73

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -281,7 +281,7 @@ func executeHelp(p *Plugin, c *plugin.Context, header *model.CommandArgs, args .
 }
 
 func executeExport(p *Plugin, c *plugin.Context, header *model.CommandArgs, args ...string) *model.CommandResponse {
-	config, err := json.MarshalIndent(p.API.GetConfig().PluginSettings.Plugins["mattermost-autolink"], "", "    ")
+	config, err := json.MarshalIndent(p.getConfig(), "", "    ")
 
 	if err != nil {
 		return responsef(err.Error())
--- a/server/command.go
+++ b/server/command.go
@@ -330,7 +330,7 @@ func executeExport(p *Plugin, c *plugin.Context, header *model.CommandArgs, args
 		return responsef(postErr.Error())
 	}
 
-	return responsef("Exported `autolink-config.json`")
+	return responsef("Exported `" + file.Name + "`")
 }
 
 func responsef(format string, args ...interface{}) *model.CommandResponse {
--- a/server/command.go
+++ b/server/command.go
@@ -284,21 +284,23 @@ func executeHelp(p *Plugin, c *plugin.Context, header *model.CommandArgs, args .
 }
 
 func executeImport(p *Plugin, c *plugin.Context, header *model.CommandArgs, args ...string) *model.CommandResponse {
-	posts, postsErr := p.API.GetPostsForChannel(header.ChannelId, 0, 1)
-	if postsErr != nil {
-		return responsef(postsErr.Error())
+	posts, err := p.API.GetPostsForChannel(header.ChannelId, 0, 1)
+	if err != nil {
+		return responsef(err.Error())
 	}
 
 	fileID := posts.ToSlice()[0].FileIds[0]
 
-	file, fileErr := p.API.GetFile(fileID)
-	if fileErr != nil {
-		return responsef(fileErr.Error())
+	var file []byte
+	file, err = p.API.GetFile(fileID)
+	if err != nil {
+		return responsef(err.Error())
 	}
 
-	info, infoErr := p.API.GetFileInfo(fileID)
-	if infoErr != nil {
-		return responsef(infoErr.Error())
+	var info *model.FileInfo
+	info, err = p.API.GetFileInfo(fileID)
+	if err != nil {
+		return responsef(err.Error())
 	}
 
 	var config Config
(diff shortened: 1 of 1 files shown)


adarj:
Screenshot of the import/export functionality:

Screenshot from 2019-09-04 19-16-13

server/command.go
```diff
@@ -278,6 +283,55 @@ func executeHelp(p *Plugin, c *plugin.Context, header *model.CommandArgs, args .
 	return responsef(helpText)
 }
 
+func executeImport(p *Plugin, c *plugin.Context, header *model.CommandArgs, args ...string) *model.CommandResponse {
```
levb: 0/5 We may want to support 2 kinds of "import": **replace** which is what you have here, and **update** when the new config is applied as an update, matching on Name (Pattern as a secondary?). The matched entries are updated, the unmatched ones are added?

cc @lieut-data for an opinion - I think he might advocate sticking with the idempotent **replace** only, as is. For "manual" use by a sysadmin, having an update makes a lot of sense.

One reason to have the "update" mode would be that we can then have library of snippets in this repo, that customers can customize and upload one at a time. Can add a link to the plugin's config page :)

server/command.go
```diff
@@ -278,6 +283,55 @@ func executeHelp(p *Plugin, c *plugin.Context, header *model.CommandArgs, args .
...
+	posts, err := p.API.GetPostsForChannel(header.ChannelId, 0, 1)
+	if err != nil {
+		return responsef(err.Error())
+	}
+
```
levb: 2/5 It is my understanding that `GetPostsForChannel` may return a `nil` error, and yet an empty result set. We should check for it, and return a meaningful error message, explaining the usage (last post should have the attachment to import)

server/command.go
```diff
@@ -278,6 +283,55 @@ func executeHelp(p *Plugin, c *plugin.Context, header *model.CommandArgs, args .
...
+	var info *model.FileInfo
+	info, err = p.API.GetFileInfo(fileID)
+	if err != nil {
+		return responsef(err.Error())
+	}
+
+	var config Config
+	configErr := json.Unmarshal(file, &config)
+	if configErr != nil {
+		return responsef(configErr.Error())
```
levb: 2/5 Should wrap the error with a user-friendly error message, explaining the expected input format

server/command.go
```diff
@@ -278,6 +283,55 @@ func executeHelp(p *Plugin, c *plugin.Context, header *model.CommandArgs, args .
...
+	}
+
```
levb: 0/5 maybe add an extra check that the result of the decode is not empty, effectively checking against invalid inputs that are valid JSON.

server/command.go
```diff
@@ -278,6 +283,55 @@ func executeHelp(p *Plugin, c *plugin.Context, header *model.CommandArgs, args .
...
+	return responsef("Imported `" + info.Name + "`")
+}
+
+func executeExport(p *Plugin, c *plugin.Context, header *model.CommandArgs, args ...string) *model.CommandResponse {
+	config, _ := json.MarshalIndent(p.getConfig(), "", "    ")
+	info, _ := p.API.UploadFile(config, header.ChannelId, "autolink-config.json")
+
+	_, err := p.API.CreatePost(&model.Post{
+		UserId:    header.UserId,
```
levb: Should probably add message text here

server/command.go
adarj: Would the update functionality be implemented as a separate function (i.e. `executeUpdate`), or as a flag that gets checked in the `executeImport` function (i.e. `if isUpdate == true {...}`)?

levb: after you said it, I think 2 separate commands, `/autolink replace` and `/autolink update` would make most sense.

server/command.go
```diff
@@ -278,6 +283,55 @@ func executeHelp(p *Plugin, c *plugin.Context, header *model.CommandArgs, args .
 	return responsef(helpText)
 }
 
+func executeImport(p *Plugin, c *plugin.Context, header *model.CommandArgs, args ...string) *model.CommandResponse {
+	posts, err := p.API.GetPostsForChannel(header.ChannelId, 0, 1)
```
crspeller: `header` is a bit confusing name for the command args. Generally we use `args` or `commandArgs`.

server/command.go
```diff
@@ -278,6 +283,55 @@ func executeHelp(p *Plugin, c *plugin.Context, header *model.CommandArgs, args .
...
+
+	p.updateConfig(func(conf *Config) {
+		*conf = config
+	})
+
+	return responsef("Imported `" + info.Name + "`")
+}
+
+func executeExport(p *Plugin, c *plugin.Context, header *model.CommandArgs, args ...string) *model.CommandResponse {
+	config, _ := json.MarshalIndent(p.getConfig(), "", "    ")
```
crspeller: We should handle these errors.

hanzei:
@adarj Do you have questions about the feedback?

levb:
@adarj are you still interested in finishing this ticket? Anything we can help with?

hanzei:
Closing this one due to inactivity.

PR #76, opened 2019-09-08
Hashtags Inserted by Autolink Searchable

### LINKED ISSUE #20: Hashtags inserted by autolink do not work
I wanted to use autolink to also tag the links by doing something like `[src](url "#tag")`, which is then picked up by the tag search.

Unfortunately autolink plugin v0.2 does not work correctly with hashtags.

To reproduce use the following substitution:
```json
"Pattern": "foo",
"Template": "#bar"
```
Then search for `#bar` and verify that the search does not find any results.

Editing the message after the substitution has occurred does produce the desired result.

mjthomp95:
Enables Autolink inserted hashtags to be searched for.

Fixes #20

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -24,4 +24,8 @@ require (
 )
 
 // Workaround for https://github.com/golang/go/issues/30831 and fallout.
-replace github.com/golang/lint => github.com/golang/lint v0.0.0-20190227174305-8f45f776aaf1
+replace (
+	git.apache.org/thrift.git => github.com/apache/thrift v0.0.0-20180902110319-2566ecd5d999
+	// Workaround for https://github.com/golang/go/issues/30831 and fallout.
+	github.com/golang/lint => github.com/golang/lint v0.0.0-20190227174305-8f45f776aaf1
+)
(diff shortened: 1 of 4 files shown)


mjthomp95:
The circleci failed because of not being able to connect to git.apache.org/thrift, which is moved to github.com/apache/thrift. This [issue](https://github.com/kubernetes/minikube/issues/5245) provides a solution should I include this in my PR or should this be opened into a different issue?

hanzei:
Please include
```
replace (
	git.apache.org/thrift.git => github.com/apache/thrift v0.0.0-20180902110319-2566ecd5d999
	// Workaround for https://github.com/golang/go/issues/30831 and fallout.
	github.com/golang/lint => github.com/golang/lint v0.0.0-20190227174305-8f45f776aaf1
)
```

into your PR

lindalumitchell:
@lieut-data or @levb can/should there be a JIRA ticket for this change?

(Thanks for already asking about unit tests, and thanks @mjthomp95 for adding them. :) )

lieut-data:
@lindalumitchell, I'll defer to @levb on that for how we track work done in plugins. I hope one day we can find a way to avoid duplicating every GitHub issue in JIRA.

levb:
@dylan I am going to cut an Autolink release soon, let's test it there

PR #77, opened 2019-09-29
Bumped version to 1.1.1, requires 5.16+

levb:


plugin.json
```diff
@@ -2,7 +2,8 @@
   "id": "mattermost-autolink",
   "name": "Autolink",
   "description": "Automatically rewrite text matching a regular expression into a Markdown link.",
-  "version": "1.1.0",
+  "version": "1.1.1",
+  "min_server_version": "5.15.0",
```
hanzei: Out of interest: Why is this needed?

levb: /autolink command saves the config, causing OnConfigChanged. In older versions and older bundled plugins there were known deadlocks. We really should be requiring 5.16 now where the deadlocks and other related bugs are fixed, but we can't yet.

levb:
> You need to bump the version in the generated files

Right, I _chose_ not to do it since they are ignored and re-generated by the build. Is there a reason that the committed versions of the manifests to be up to date with plugin.json?

plugin.json
hanzei: Why can't we requiring 5.16?

hanzei:
> > You need to bump the version in the generated files
> 
> Right, I _chose_ not to do it since they are ignored and re-generated by the build. Is there a reason that the committed versions of the manifests to be up to date with plugin.json?

To avoid confusion. The next time im developing I will have to unrelated changes in my branch, that I can't get rid of. Plus the plugins might rely on the version information.

plugin.json
levb: Because chicken and egg? If we publish this release of Autolink now, it'd be requiring a non-existing Mattermost version for 2 weeks?

hanzei: The required server version is compared to `model.CurrentVersion`, which was bumped in https://github.com/mattermost/mattermost-server/pull/12172. We can safely bump the version here.

Please also note that `master` server is using `5.18`, hence you can bump the required version to `5.18`, if your plugin relies on something that needs 5.18.

levb:
Will up the source files.

OT perhaps, but I'd advocate for removing these files completely. They increase confusion by duplicating the data out of plugin.json, and are 100% generated.

For golang, we can accomplish this by setting existing global variables from the Makefile. For JavaScript I am sure we can do similar. Then the committed "in-code" version/ID values  can be anything at all ("dev"?), and only changed at build time.

plugin.json
levb: Right, but none of these are _released_ and available for download by the customers, yet. Right?

hanzei: That is correct. I think it comes down to what "requires" means. If you highly encourage people to use `5.16` because of bug fixes, do you want to enforce this?

I'm 1/5 that it's fine to release a plugin for a non existing release. I suppose people won't jump eminently on this new release and install it.

levb: Alright, I'll go with it, particularly because 5.16 is the intent here. With the advent of the plugin marketplace I suppose it'll becomes less relevant of an issue going forward since most customers will be installing from there, and presumably we wouldn't serve an incompatible version anyway.

PR #79, opened 2019-11-15
Add enable on update (fixes #36)

### LINKED ISSUE #36: Updated posts are not processed

grubbins:
Config item `EnableOnUpdate` allows processing on updated posts too.

### DIFF
--- a/server/plugin_test.go
+++ b/server/plugin_test.go
@@ -77,6 +77,7 @@ func TestSpecialCases(t *testing.T) {
 	})
 	validConfig := Config{
 		EnableAdminCommand: false,
+		EnableOnUpdate:     true,
 		Links:              links,
 	}
 
@@ -220,13 +221,54 @@ func TestSpecialCases(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.inputMessage, func(t *testing.T) {
-			post := &model.Post{
-				Message: tt.inputMessage,
+			{
+				// user creates a new post
+
+				post := &model.Post{
+					Message: tt.inputMessage,
+				}
+
+				rpost, _ := p.MessageWillBePosted(&plugin.Context{}, post)
+
+				assert.Equal(t, tt.expectedMessage, rpost.Message)
 			}
+			{
+				// user updates the modified post but with no changes
 
-			rpost, _ := p.MessageWillBePosted(&plugin.Context{}, post)
+				post := &model.Post{
+					Message: tt.expectedMessage,
+				}
 
-			assert.Equal(t, tt.expectedMessage, rpost.Message)
+				rpost, _ := p.MessageWillBeUpdated(&plugin.Context{}, post, post)
+
+				assert.Equal(t, tt.expectedMessage, rpost.Message)
+			}
+			{
+				// user updates the modified post and sets it back to the original text
+
+				originalPost := &model.Post{
+					Message: tt.expectedMessage,
+				}
+				post := &model.Post{
+					Message: tt.inputMessage,
+				}
+
+				rpost, _ := p.MessageWillBeUpdated(&plugin.Context{}, originalPost, post)
+
+				assert.Equal(t, tt.expectedMessage, rpost.Message)
+			}
+			{
+				// user updates an empty post to the original text
+
+				emptyPost := &model.Post{}
+				post := &model.Post{
+					Message: tt.inputMessage,
+				}
+
+				rpost, _ := p.MessageWillBeUpdated(&plugin.Context{}, post, emptyPost)
+
+				assert.Equal(t, tt.expectedMessage, rpost.Message)
+			}
 		})
 	}
 }
--- a/plugin.json
+++ b/plugin.json
@@ -19,6 +19,12 @@
         "display_name": "Enable administration with /autolink command",
         "type": "bool",
         "default": false
+      },
+      {
+        "key": "EnableOnUpdate",
+        "display_name": "Apply plugin to updated posts as well as new posts",
+        "type": "bool",
+        "default": false
       }
     ]
   }
--- a/server/config.go
+++ b/server/config.go
@@ -12,6 +12,7 @@ import (
 // Config from config.json
 type Config struct {
 	EnableAdminCommand bool
+	EnableOnUpdate     bool
 	Links              []Link
 }
 
@@ -74,6 +75,7 @@ func (conf Config) ToConfig() map[string]interface{} {
 	}
 	return map[string]interface{}{
 		"EnableAdminCommand": conf.EnableAdminCommand,
+		"EnableOnUpdate":     conf.EnableOnUpdate,
 		"Links":              links,
 	}
 }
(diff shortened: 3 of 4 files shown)


levb:
This will need to merge https://github.com/mattermost/mattermost-plugin-autolink/pull/80 to build

grubbins:
I'll look at a test case, yes.

I am working on getting the CLA signed...

levb:
@grubbins were you able to sign the CLA?

jasonblais:
Hi @grubbins, let us know if you have any questions about the CLA? Happy to help clarify anything

grubbins:
Sorry for the delay all - the CLA will be fine but I need to get it signed at my employer and then resubmit this change under a new account that's a member of their organisation account.

jasonblais:
@grubbins no worries, thanks for the heads up!

grubbins:
Added this to `plugin_test`, and CLA finally in  place.

grubbins:
Actually no. The test is ineffective. Give me a minute...

grubbins:
OK now the test actually does fail if I remove the update handling :)

I did not explicitly test that the plugin does *not* modify an updated post if the config item is disabled (although I did see that behaviour).

server/plugin_test.go
```diff
@@ -220,13 +221,54 @@ func TestSpecialCases(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.inputMessage, func(t *testing.T) {
-			post := &model.Post{
-				Message: tt.inputMessage,
+			{
```
levb: LOL, that's an interesting way of formatting! 0/5 a slight preference to make each "block" a separate t.Run, and use the test names for comments - makes the logs pretty?

grubbins: Yeah I don't have a lot of experience structuring go tests, but I also felt the whole test could probably be re-arranged to more clearly report which test failed. I decided that would be better tackled by others...

levb: Ok, I'll do it later.

levb:
@DHaussermann ready for you to test

DHaussermann:
@grubbins Thanks for this improvement! It is functional. When I update a post the pattern is applied as expected. 
The one concern I have is that when I tested it locally I did not see the line `"enableonupdate": ` appear in my config. I had to manually add the line to test this change. Is it possible this is an issue with the code change and not something I'm doing? Is this something you can reproduce?

grubbins:
> when I tested it locally I did not see the line `"enableonupdate": ` appear in my config when I uploaded the build

I am not sure what's expected here actually. I would not expect the config file to be updated when a plugin is uploaded. Maybe next time mattermost saves its config (i.e. exits, or you hit a save button in system console) it would write all the config fields?

server/plugin_test.go
grubbins: BTW in case it's not obvious, the separate blocks are there to ensure we destroy the `post` instances because the plugin mutates the ones passed in... and it seemed even more messy to declare them all with separate names.

grubbins:
@DHaussermann I just tried uploading the plugin to our test server (over-writing existing). There is no config section for this plugin in `config.json` and it does not seem to automatically create anything. I believe you just have to configure this file manually.

DHaussermann:
Strange. I believe the line above it for `enableadmincommand` appeared in the config once a version that includes this is uploaded. So, I was expecting the new line to appear with a value of false that I could go change.  
Screen Shot 2020-01-21 at 9 49 17 PM

@levb do you have any thoughts on this?

grubbins:
@DHaussermann I believe that is because the `enableadmincommand` configuration is controllable via the System Console, therefore mattermost can write it to `config.json` - but none of the other configuration of this plugin is available via the System Console.

levb:
@grubbins since you implemented the config setting as a global flag, perhaps it is indeed worth adding it to `plugin.json` so it is a visible setting? I'm 0/5 on this.

Also, this made me think that perhaps it should have been implemented as a per-link setting, rather than a global one? I hate to go back and re-do it but it really does seem more appropriate, what do you think?

grubbins:
I could look at adding it to the `plugin.json` yes!

I don't want to make it per-link. I added this functionality in the first case because my users were very confused that the behaviour works on new posts but not updates (and assumed this was a bug), and everyone asked for it to work on update; I think it would be even *more* confusing if this happened for some link rules and not others.

Conversely the argument for leaving this feature turned off, if I understand rightly, is that it gives users a way to undo/suppress the autolink action. In that case I guess the users on that mattermost instance would need to be told this, so they can make use of it. Again, I think it would then be *more* confusing for them if this worked for some link rules and not others. (I also strongly feel that if we want an intuitive way for users to suppress the autolinking, turning off enable on update is not the solution)

So I would really prefer to get this merged, and if the need for per-link setting is really there then this global flag can act as a default and we could later add over-rides per link - how about that?

DHaussermann:
@grubbins the change implemented is working. As it is now, this is already very useful! I have no issue with addressing the follow-up concerns in a separate issue. However, since the flag is not exposed in the UI, we would at least need to update the read-me to include this option (I did not see a change for this).

cc @aaronrothschild ☝️.

grubbins:
@DHaussermann  agreed - I pushed another version which has the new setting in `plugin.json`. I hope that's enough for it to appear in System Console but I did not have time to test that yet...

grubbins:
OK, I tested it again with the entry in `plugin.json` and indeed the new option is now visible via System Console. It is not immediately written to `config.json`, but as expected if I save the config in System Console then the new configuration item is written to `config.json`.

levb:
@grubbins I am convinced! Thanks for the detailed reasoning!

levb:
@DHaussermann I think we just need your ok on the newly visible flag in the UI.

aaronrothschild:
@levb @grubbins I got this comment from a customer - thoughts on this scenario?  Separate PR (I know this is merged already)?

> > 1. Autolink- a recent (January) community contribution will actually fix this issue- https://github.com/mattermost/mattermost-plugin-autolink/pull/79 . I talked to the PM for integrations and he’s working on getting an update/release scheduled for auto link so that we can include this. Timing TBD but hopefully soon. I’ll keep an eye on it. 

> Interesting, thanks for pointing me at the pull request. It looks like it will allow an individual auto-link rule to indicate that it should be re-run on edit, and assumes that most auto-link rules are idempotent. I'm not sure if mine are, but perhaps I can make them work correctly on re-run. 

> However, even if I can make it work I'm not sure it's exactly what users would expect. For example, suppose I have a ticketing system auto-link that converts something like XYZ-5432 into [XYZ-5432](http://my.fake.ticketing.system/XYZ-5432). If I meant to link to XYZ-54321, then when I edit I need to modify the link and the link text. Ideally, editing a post that's been auto-linked would show up as the pre-auto-linked text ("XYZ-5432") when editing and then have the auto-link re-applied.

levb:
@aaronrothschild - I think it is a new ticket, and maybe something for the platform team to consider  (versioning the post, in its most general manifestation).

Issue #84, opened 2020-01-21
Add the concept of plugin admin users


levb:
Add a (visible, editable in admin console) section to autolink config to define the list of plugin administrators. It should take a list of @usernames, ideally with autocomplete.

These users should then be allowed to execute the `/autolink` command.

An rough prototype can be found in 
- https://github.com/mattermost/mattermost-plugin-solar-lottery/blob/e41d9d22ad8f4ebd5164631b99441ad3548ea4ab/plugin.json#L20-L25
- https://github.com/mattermost/mattermost-plugin-solar-lottery/blob/dev/server/plugin/api.go#L21

But it lacks accepting @usernames with autocomplete.

![image](https://user-images.githubusercontent.com/1187448/72811054-b6f86a00-3c13-11ea-8b9d-2163d4467b40.png)

vespian:
@levb Basing on [1], by default there is a `username` type of the field with autocompletion, but it is meant to choose one user, not a list. If one wants to have multi-user support, a custom field needs to be created, but this requires some extra web code, just like in e.g. [2]. Right? 


[1] https://developers.mattermost.com/extend/plugins/manifest-reference/#settings_schema.settings.type
[2] https://github.com/mattermost/mattermost-plugin-custom-attributes/blob/14027463acf04ecea5a0d5f2b3abfe80959f3a80/webapp/src/index.js#L12

levb:
@vespian You're right, we do not have the appropriate input type. I do not believe that investing in custom UI work is worth it in this case. How about we allow a ` `- or `,`-separated list of strings, where each one can be a mattermost user ID, or a `[@]username`?

vespian:
> How about we allow a - or ,-separated list of strings, where each one can be a mattermost user ID, or a [@]username?

+1

vespian:
I would like to give it a try.

vespian:
@levb Please have a look at https://github.com/mattermost/mattermost-plugin-autolink/pull/93

PR #93, opened 2020-02-16
[GH-84] Add the concept of plugin admins

vespian:
#### Summary
This plugin adds the concept of plugin admin users. A new plugin configuratoin option was added which lists comma-separated usernames for which modifying the configuration of the plugin is allowed (no auto-compleate, see [3]). Please see below for the sample config window:

autolink_preview

and sample interaction with the plugin as a user who first did not have admin privileges, and then was granted ones:

autolink-error-non-admin

While working on this change I have also changed how the `Config` is handled/passed around. In Go, function arguments are passed by value [1], so if. e.g. the underlying array of the `Config.Links` changed (i.e. new entries were added, causing reallocation), the other copies would be still using the old underlying array which could lead to weird errors. Also, based on the plugin template[2], the thread-safety seems to be guaranteed by passing around the reference to the plugin's config, so that if one goroutine updates the config, the other still has the consistent view on the old configuration as it is holding the old reference. Happy to discuss!

Do we still need `EnableAdminCommand` config option in this plugin, now that we have plugin administrators? Do we want to allow also adding userIds in the new field?

[1] https://stackoverflow.com/a/47486724
[2] https://github.com/mattermost/mattermost-plugin-starter-template/blob/282a8cf227849a25e306caff11597a0e4ba4cd01/server/configuration.go
[3] https://github.com/mattermost/mattermost-plugin-autolink/issues/84#issuecomment-584928588

#### Ticket Link
  Fixes ~~https://github.com/mattermost/mattermost-server/issues/84~~ 
  Fixes https://github.com/mattermost/mattermost-plugin-autolink/issues/84  (jfrerich 2/28/20)

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -32,7 +32,7 @@
         "key": "PluginAdmins",
         "display_name": "Admin User IDs",
         "type": "text",
-        "help_text": "Comma-separated list of userIds of users authorized to administer the plugin in addition to the System Admins.\n \nUser IDs can be found by navigating to **System Console** > **Users**. After clicking into a user's name, their ID is on the right-hand side of the blue header."
+        "help_text": "Comma-separated list of userIDs authorized to administer the plugin in addition to the System Admins.\n \nUser IDs can be found by navigating to **System Console** > **Users**."
       }
     ]
   }
--- a/plugin.json
+++ b/plugin.json
@@ -27,6 +27,12 @@
         "display_name": "Apply plugin to updated posts as well as new posts",
         "type": "bool",
         "default": false
+      },
+      {
+        "key": "PluginAdmins",
+        "display_name": "Admin User IDs",
+        "type": "text",
+        "help_text": "Comma-separated list of userIds of users authorized to administer the plugin in addition to the System Admins.\n \nUser IDs can be found by navigating to **System Console** > **Users**. After clicking into a user's name, their ID is on the right-hand side of the blue header."
       }
     ]
   }
--- a/server/autolinkplugin/config.go
+++ b/server/autolinkplugin/config.go
@@ -14,23 +14,33 @@ import (
 type Config struct {
 	EnableAdminCommand bool
 	EnableOnUpdate     bool
+	PluginAdmins       string
 	Links              []autolink.Autolink
+
+	// AdminUserIds is a set of UserIds that are permitted to perform
+	// administrative operations on the plugin configuration (i.e. plugin
+	// admins). On each configuration change the contents of PluginAdmins
+	// config field is parsed into this field.
+	AdminUserIds map[string]struct{}
 }
 
 // OnConfigurationChange is invoked when configuration changes may have been made.
 func (p *Plugin) OnConfigurationChange() error {
 	var c Config
-	err := p.API.LoadPluginConfiguration(&c)
-	if err != nil {
-		return err
+	if err := p.API.LoadPluginConfiguration(&c); err != nil {
+		return fmt.Errorf("failed to load configuration: %w", err)
 	}
 
 	for i := range c.Links {
-		err = c.Links[i].Compile()
-		if err != nil {
+		if err := c.Links[i].Compile(); err != nil {
 			mlog.Error(fmt.Sprintf("Error creating autolinker: %+v: %v", c.Links[i], err))
 		}
 	}
+
+	// NOTE(prozlach): Plugin admin UserId parsing and validation errors are
+	// not fatal, if everything fails only sysadmin will be able to manage the
+	// config which is still OK
+	c.parsePluginAdminList(p)
 
 	p.UpdateConfig(func(conf *Config) {
 		*conf = c
@@ -54,15 +64,17 @@ func (p *Plugin) OnConfigurationChange() error {
 	return nil
 }
 
-func (p *Plugin) getConfig() Config {
+func (p *Plugin) getConfig() *Config {
 	p.confLock.RLock()
 	defer p.confLock.RUnlock()
+
 	return p.conf
 }
 
 func (p *Plugin) GetLinks() []autolink.Autolink {
 	p.confLock.RLock()
 	defer p.confLock.RUnlock()
+
 	return p.conf.Links
 }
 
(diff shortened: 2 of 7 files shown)


server/autolinkplugin/config.go
```diff
@@ -77,34 +89,61 @@ func (p *Plugin) SaveLinks(links []autolink.Autolink) error {
 	return nil
 }
 
-func (p *Plugin) UpdateConfig(f func(conf *Config)) Config {
```
vespian: The return value was never used.

server/autolinkplugin/plugin_test.go
```diff
@@ -55,6 +57,259 @@ func TestPlugin(t *testing.T) {
 	assert.Equal(t, "Welcome to [Mattermost](https://mattermost.com)!", rpost.Message)
 }
 
+type SuiteAuthorization struct {
```
vespian: I took the liberty of not being consistent with other tests in this file, as using testing suite approach IMO is cleaner and we are already using testing suites in e.g. `mattermost-server` repo.

plugin.json
```diff
@@ -27,6 +27,12 @@
         "display_name": "Apply plugin to updated posts as well as new posts",
         "type": "bool",
         "default": false
+      },
+      {
+        "key": "AdminUsernames",
```
crspeller: Generally we only accept user IDs in fields like this. Since typos can happen and usernames can be changed, it is a minor security risk. 

We hope to fix the UX with a user picker component that would ease the process for admins. 

Do you mind updating this to use IDs instead of names?

vespian:
@crspeller Done, ready for another pass.

server/autolinkplugin/config.go
```diff
@@ -14,24 +14,34 @@ import (
...
 
 	for i := range c.Links {
-		err = c.Links[i].Compile()
-		if err != nil {
+		if err := c.Links[i].Compile(); err != nil {
 			mlog.Error(fmt.Sprintf("Error creating autolinker: %+v: %v", c.Links[i], err))
 		}
 	}
 
+	// NOTE(prozlach): Plugin admin UserId parsing and validation errors are
```
jfrerich:
```suggestion
	// Plugin admin UserId parsing and validation errors are
```

server/autolinkplugin/config.go
```diff
@@ -77,34 +89,58 @@ func (p *Plugin) SaveLinks(links []autolink.Autolink) error {
...
+// parsePluginAdminList parses the contents of PluginAdmins config field
+func (conf *Config) parsePluginAdminList(p *Plugin) {
+	conf.AdminUserIds = make(map[string]struct{})
+
+	if len(conf.PluginAdmins) == 0 {
+		// There were no plugin admin users defined
+		return
+	}
+
+	split := strings.Split(conf.PluginAdmins, ",")
```
jfrerich: Until we implement user names, I suggest we s/split/userIDs/

vespian:
@jfrerich Done, changes in commit `Review fixes #1`.

server/autolinkplugin/config.go
```diff
@@ -14,24 +14,34 @@ import (
...
 	EnableAdminCommand bool
 	EnableOnUpdate     bool
+	PluginAdmins       string
 	Links              []autolink.Autolink
+
+	// AdminUserIds is a set of UserIds that are permitted to perform
+	// administrative operations on the plugin configuration (i.e. plugin
+	// admins). On each configuration change the contents of PluginAdmins
+	// config field is parsed into this field.
+	AdminUserIds map[string]struct{}
```
crspeller: Since the number of admin users is likely to be low, I would maybe skip the complexity here and just have the check use the string directly instead of pre-processing it. eg: https://github.com/mattermost/mattermost-plugin-aws-SNS/blob/master/server/plugin.go#L348

vespian: Pre-parsing UserIds on config change [1] allows us to catch malformed UserIds and log them so that users can debug situations where e.g. `X` does not have access without resorting to digging into the code. Additionally the `map[string]struct{}` + `if _, ok := foo[x]; !ok` is a common idiom in Go for a set and set membership tests, at least I find it pretty readable.

Let me know if you feel strongly about it and I will change it as requested, if not - it would be great if we could leave it as-is.

[1] https://github.com/mattermost/mattermost-plugin-autolink/pull/93/files/bb32075fc5b2f4e01ff880d55e752741bdb2ef0b#diff-45a5ec4d780fb965962fe72d26e03952R139-R144

crspeller: Ok. Sounds good.

crspeller:
@DHaussermann @aaronrothschild

vespian:
@DHaussermann @aaronrothschild Ping

crspeller:
@vespian I pinged them on the community server. Sorry for the delay!

aaronrothschild:
Sorry for the delay on this @vespian , thanks for your contribution!

aaronrothschild:
Sorry @vespian is this a general addition for "Admins" on "Plugins" or just for the Autolink plugin?  

Just want to make sure I'm understanding correctly.  

If we have a general purpose "Admin Users" for plugins....we need to think this through a bit more.  It likely would need to be limited to Enterprise Edition licences (ie: Only EE licenced MM servers would have a centralized plce for admins of plugins).

vespian:
> Sorry @vespian is this a general addition for "Admins" on "Plugins" or just for the Autolink plugin?

@aaronrothschild This is only for autolink plugin :)

DHaussermann:
@vespian  I'm seeing an odd bug here. The list of authorized user is clearing itself out and it seem related maybe to when the functionality is used to change autolink settings. It seems like it only works once for a non-admin user and then the list empties. 

Steps:
- Login as sys admin
- Pick a target non-system admin user and note the ID
- Open autolink config and save the ID as an authorized user (optionally, can also add other data)
- Login in a separate browser session as target user
- Use commands like `list` and `help` repeatedly and not that it keeps working
- Use a command that changes the data such as `enable`, `disable` or set `WordMatch` 
- Try using any other autolink command
**Observed**
- Users see's message about functionality being locked to sysadmin
- Return to admin browser session and refresh the list is now empty

Can you please see if you can also repro this? Please reach out if there is anymore clarification or details I can provide.

vespian:
@DHaussermann Thanks a lot for checking my work, and a detailed bug report! I will take a closer look into it during the weekend, maybe earlier but no promises.

vespian:
@DHaussermann Fixed in commit https://github.com/mattermost/mattermost-plugin-autolink/pull/93/commits/daefa31eeda692dcc6d5330a1fd7afd45647fd7c 

Let me know if you find anything more. Thanks!

Issue #94, opened 2020-01-24
Autolink should ignore bot posts. (was: Commit notifications trigger unexpected link preview)


lieut-data:
To avoid clashes with bot-formatted links, Autolink should (optionally?) ignore bot posts.

1/5 recommend adding an "Process posts from Bots", default `false` to the plugin config.

---
Original description:

I recently received a notification in the [message of a commit](https://github.com/mattermost/mattermost-sa-code/commit/528462af4dca50e41011fe1076c76edb00327b5b):

![image](https://user-images.githubusercontent.com/1023171/73068914-ca5d3c80-3e82-11ea-9a13-39edad97c9ab.png)

In my summary, this then resulted in a perhaps unexpected link preview compared to all the other links in the message:

![image](https://user-images.githubusercontent.com/1023171/73068947-df39d000-3e82-11ea-8ccb-25bf3b8fd795.png)

levb:
I think the issue should stay here, and be repurposed if the history is important. Autolink will already expose the link config APIs in the next release, no special work needed there AFAIU. We just need to add the code on the GiitHub side to manage the links, and probably a config/feature flag to use this new capability instead of the current in-plugin substitution. cc @jfrerich

levb:
(closed by mistake, reopened)

lieut-data:
I think there's two separate issues here:
* autolink shouldn't process messages by bots (open for discussion), which I think is the meat of this issue
* GitHub should rely on autolink for link substituion (not as clear to me, and also open for discussion)

One reason to keep this in the GitHub plugin is to solve the second issue, which I guess also solves the first issue? One concern is that this means the GitHub plugin, out of the box, isn't as fully-featured as GitHub + AutoLink, and I have no idea of the lifecycle of these plugin-to-plugin registrations. Seems like a thorny problem and worthy of design + investigation, as compared to solving the original report.

Thoughts on this interpretation?

levb:
> autolink shouldn't process messages by bots (open for discussion)

@lieut-data I'd argue against this complexity. Autolink should already ignore substituting _links_ in markdown; Bots should be aware of the markdown nature of the contents and properly encode links in the message bodies if they come from external systems as raw text or other formats. Ditto for the messages that the bots generate.

> GitHub should rely on autolink for link substitution (not as clear to me, and also open for discussion)

2 motives I can think of:
[...]

PR #98, opened 2020-03-07
[GH-84] Do not rewrite messages from bots

vespian:
#### Summary

This PR makes autolink plugin not rewrite messages from other bots as discussed in https://github.com/mattermost/mattermost-plugin-autolink/issues/94

#### Ticket Link

  Fixes https://github.com/mattermost/mattermost-plugin-autolink/issues/94

### DIFF
--- a/server/autolinkplugin/plugin.go
+++ b/server/autolinkplugin/plugin.go
@@ -78,7 +78,29 @@ func contains(team string, channel string, list []string) bool {
 	return false
 }
 
+func (p *Plugin) isBotUser(userId string) (bool, *model.AppError) {
+	user, appErr := p.API.GetUser(userId)
+	if appErr != nil {
+		p.API.LogError("failed to check if message for rewriting was send by a bot", "error", appErr)
+		return false, appErr
+	}
+
+	return user.IsBot, nil
+}
+
 func (p *Plugin) ProcessPost(c *plugin.Context, post *model.Post) (*model.Post, string) {
+	isBot, appErr := p.isBotUser(post.UserId)
+	if appErr != nil {
+		// NOTE: Not sure how we want to handle errors here, we can either:
+		// * assume that occasional rewrites of Bot messges are ok
+		// * assume that occasional not rewriting of all messages is ok
+		// Let's assume for now that former is a lesser evil and carry on.
+	}
+	if isBot {
+		p.API.LogDebug("not rewriting message from bot", "userId", post.UserId)
+		return nil, ""
+	}
+
 	conf := p.getConfig()
 	postText := post.Message
 	offset := 0
--- a/server/autolinkplugin/plugin.go
+++ b/server/autolinkplugin/plugin.go
@@ -89,18 +89,6 @@ func (p *Plugin) isBotUser(userId string) (bool, *model.AppError) {
 }
 
 func (p *Plugin) ProcessPost(c *plugin.Context, post *model.Post) (*model.Post, string) {
-	isBot, appErr := p.isBotUser(post.UserId)
-	if appErr != nil {
-		// NOTE: Not sure how we want to handle errors here, we can either:
-		// * assume that occasional rewrites of Bot messges are ok
-		// * assume that occasional not rewriting of all messages is ok
-		// Let's assume for now that former is a lesser evil and carry on.
-	}
-	if isBot {
-		p.API.LogDebug("not rewriting message from bot", "userId", post.UserId)
-		return nil, ""
-	}
-
 	conf := p.getConfig()
 	postText := post.Message
 	offset := 0
@@ -170,7 +158,23 @@ func (p *Plugin) ProcessPost(c *plugin.Context, post *model.Post) (*model.Post, 
 
 		return true
 	})
-	post.Message = postText
+	if post.Message != postText {
+		isBot, appErr := p.isBotUser(post.UserId)
+		if appErr != nil {
+			// NOTE: Not sure how we want to handle errors here, we can either:
+			// * assume that occasional rewrites of Bot messges are ok
+			// * assume that occasional not rewriting of all messages is ok
+			// Let's assume for now that former is a lesser evil and carry on.
+		} else if isBot {
+			// We intentionally use a single if/else block so that the code is
+			// more readable and does not relly on hidden side effect of
+			// isBot==false when appErr!=nil.
+			p.API.LogDebug("not rewriting message from bot", "userId", post.UserId)
+			return nil, ""
+		}
+
+		post.Message = postText
+	}
 
 	post.Hashtags, _ = model.ParseHashtags(post.Message)
 
--- a/server/autolink/autolink_test.go
+++ b/server/autolink/autolink_test.go
@@ -25,6 +25,12 @@ func setupTestPlugin(t *testing.T, l autolink.Autolink) *autolinkplugin.Plugin {
 		TeamId: "theteam_id0123456789012345",
 		Name:   "thechannel_name",
 	}, (*model.AppError)(nil))
+
+	testUser := model.User{
+		IsBot: false,
+	}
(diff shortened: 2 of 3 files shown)


server/autolinkplugin/plugin.go
```diff
@@ -61,7 +62,26 @@ func contains(team string, channel string, list []string) bool {
...
 
+func (p *Plugin) isBotUser(userId string) bool {
+	user, appErr := p.API.GetUser(userId)
+	if appErr != nil {
+		// NOTE: Not sure how we want to handle errors here, we can either:
+		// * assume that occasional rewrites of Bot messges are ok
+		// * assume that occasional not rewriting of all messages is ok
+		// Let's assume for now that former is a lesser evil
+		mlog.Error("failed to check if message for rewriting was send by a bot", zap.Error(appErr))
+		return false
```
hanzei: Why are you logging the error here instead of returning it?

server/autolinkplugin/plugin.go
```diff
@@ -61,7 +62,26 @@ func contains(team string, channel string, list []string) bool {
...
+	}
+
+	return user.IsBot
+}
+
 func (p *Plugin) ProcessPost(c *plugin.Context, post *model.Post) (*model.Post, string) {
+	if p.isBotUser(post.UserId) {
+		mlog.Debug("not rewriting message from bot", zap.String("userId", post.UserId))
```
hanzei: In the future, the whole code base should use [`p.API.LogDebug`](https://developers.mattermost.com/extend/plugins/server/reference/#API.LogDebug) instead. It would be nice if you start using the with this PR. But it's not blocking.

go.mod
```diff
@@ -7,4 +7,5 @@ require (
 	github.com/mattermost/mattermost-server/v5 v5.18.0
 	github.com/pkg/errors v0.8.1
 	github.com/stretchr/testify v1.4.0
+	go.uber.org/zap v1.10.0
```
hanzei: If you use `LogDebug`, this import is not longer needed

server/autolinkplugin/plugin.go
vespian: According to [1], the only option we have while handling Post message in a hook is to either reject it, or accept it as is. Even if I return the error to the caller, there is not much we can do with it. So the idea is to simply procces posts which for which we cannot establish origin (bot/non-bot). 

Happy to discuss.


[1] https://github.com/mattermost/mattermost-server/blob/57f3ca975f565b0235d72ba6eb098e73a25513cd/plugin/hooks.go#L92-L115

vespian:
@hanzei Ready for another pass.

server/autolinkplugin/plugin.go
hanzei: My comment was more about the code flow. In go, it's usually up to the caller to decide who an error should be handled (returned, logged, ignored). Hence, my proposal would be to change the signature to ` isBotUser(userId string) (bool, *model.AppError)` and move the logic to `ProcessPost`.

vespian: Makes sense. Done.

server/autolinkplugin/plugin.go
```diff
@@ -61,7 +61,29 @@ func contains(team string, channel string, list []string) bool {
...
+	if appErr != nil {
+		p.API.LogError("failed to check if message for rewriting was send by a bot", "error", appErr)
+		return false, appErr
+	}
+
+	return user.IsBot, nil
+}
+
 func (p *Plugin) ProcessPost(c *plugin.Context, post *model.Post) (*model.Post, string) {
+	isBot, appErr := p.isBotUser(post.UserId)
```
levb: It is likely considerably more expensive to go back to the server to get the user record than first evaluating to see if any links apply. 3/5 we should move this check after an applicable link is found.

vespian: @levb I am curious - what does 3/5, 0/5 mean? I often see it in reviews.

vespian: Wrt. your comment - we do not really have data on how big/how numerous the rules can get, and how much traffic we may need to rewrite and how the plugin system and querying the server performs under load. The rewriting rules also already call `p.API.GetChannel` and `p.API.GetTeam` api calls which may be heavier than the `p.API.GetUser` that we are adding and actually having `isBot` check early may save CPU cycles.

I believe optimizing it now is a bit early. Would you mind if we ask @hanzei to add his two cents?

vespian: @levb @hanzei Ping.

levb: (waiting for @hanzei also cc @lieut-data) A single RPC/post is too expensive, anywhere in a plugin, period. We should move the RPCs to after the match, so we only execute them when needed.

Now, I see the argument that if this PR follows the existing pattern, we should stick with it as is, and have a separate ticket to optimize. I'm 1/5 that we shouldn't - let's at least make sure this PR doesn't add overhead.

https://docs.mattermost.com/process/training.html#id8

lieut-data: @levb, based on @ali-farooq0's preliminary analysis of the RPC overhead on community, it /seems/ like RPC, itself, isn't that expensive. The real cost is borne by the database queries to resolve this information, but I'm assuming we're already doing one or more of those to post the message in the first place.

That being said, no objections to moving this check below and wrapping in a `post.Message != postText` test first. Agree we should have a ticket to optimize regardless.

vespian: Thanks for the discussion! Requested changes can be found in ` Do not rewrite messgaes from bots - review fixes #1` commit.

vespian:
@levb Ready for another pass!

server/autolinkplugin/plugin.go
```diff
@@ -131,7 +141,23 @@ func (p *Plugin) ProcessPost(c *plugin.Context, post *model.Post) (*model.Post,
 
 		return true
 	})
-	post.Message = postText
+	if post.Message != postText {
+		isBot, appErr := p.isBotUser(post.UserId)
```
levb: It seems that if there is a "system failure" in the process of modifying the message, the right thing to do is to pass it through as is. 0/5 `if isBot || appErr != nil {...`

vespian: What do you mean by `in the process of modifying the message`? ` isBotUser` does not modify the message.

jfrerich:
@vespian, can you please merge from master and resolve the conflict in `plugin_test.go`

@lieut-data, gentle ping for review

lieut-data:
Thanks, @jfrerich! Happy to defer to @hanzei and @levb's reviews here -- nothing blocking from my side :)

jfrerich:
@lieut-data thanks for checking back in.  I know we already have the two dev reviews and just checking you had no further comments :)

vespian:
> @vespian, can you please merge from master and resolve the conflict in plugin_test.go

@jfrerich Done.

DHaussermann:
@vespian or @lieut-data can you help me understand the scope of this change? I'm not sure how to validate this and exactly what it should do. 

I'm seeing the the Links are still working for posts made the Jira bot or GitHub bot which is great.
When I make a post with a bot I have created the autolink is not applied even when post content matches the pattern. 
This seems like the desired behavior but how is it working? Are post from some bots already formatted to contain a link to GitHub, Jira, etc so it's not actually autolink modifying the post contents?

lieut-data:
@DHaussermann, exactly, GitHub/Jira already knows how to format its own links. The easiest way to truly verify might be to setup the conditions of https://github.com/mattermost/mattermost-plugin-autolink/issues/94 with an at-mention to yourself on a commit that autolink would have otherwise rewritten. Assuming that stops working, and the normal links stays in place, I think we're good.

lieut-data:
@aaronrothschild, any objections to this change?

PR #100, opened 2020-03-13
Add preconfigured Autolinks to plugin config

jfrerich:
### Summary
This PR adds preconfigured autolinks to the plugin system console.  A sysadmin can enable or disable each without having to configure an autolink via the slash command.

#### UI

I haven't seen a way to add a separator in the system console modal. This would be helpful to separate and label a section (ie. Preconfigured autolink section)

image

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -27,6 +27,41 @@
         "display_name": "Apply plugin to updated posts as well as new posts",
         "type": "bool",
         "default": false
+      },
+      {
+        "key": "EnableMasterCard",
+        "display_name": "MasterCard",
+        "type": "dropdown",
+        "help_text": "Conceal Mastercard Entries",
+        "default": "disabled",
+        "options": [{
+            "display_name": "disabled",
+            "value": "disabled" 
+        }, {
+            "display_name": "enabled",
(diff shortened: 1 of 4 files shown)


plugin.json
```diff
@@ -27,6 +27,41 @@
...
+        "type": "dropdown",
+        "help_text": "Conceal Mastercard Entries",
+        "default": "disabled",
+        "options": [{
+            "display_name": "disabled",
+            "value": "disabled" 
+        }, {
+            "display_name": "enabled",
+            "value": "Pattern: (?P<MasterCard](term>([Cc]ustomer [Oo]bsession)(?P<part1>5[1-5]\\d{2})[ -]?(?P<part2>\\d{4})[ -]?(?P<part3>\\d{4})[ -]?(?P<LastFour>[0-9]{4})) Template: MasterCard XXXX-XXXX-XXXX-$LastFour WordMatch: true"
+        }]
```
jfrerich: @levb please note this is a draft PR and I created it for discussion. 

This highlighted option might be ugly and would require a `value` parser, but would provide a method for adding the preconfigured autolinks to the `plugin.json` file instead of `go` code.  By adding a dropdown, we can set a value, which cannot be done with a radio (bool) type. I'm not advocating this thought it worthy of bringing up.

server/autolinkplugin/config.go
```diff
@@ -78,6 +83,35 @@ func (p *Plugin) SaveLinks(links []autolink.Autolink) error {
 	return nil
 }
 
+// GetPreConFigLinks gets the preconfigured links from the plugin config
```
jfrerich: What are your thoughts on moving this function to /autolinkplugin/preconfigured_links.go?

plugin.json
levb: I'm very confused as to why? Seems like just a boolean config option, and the go code that interprets it. _Maybe_ another checkbox to control the word boundary type.

Issue #99, opened 2020-03-09
Migrate logging to use RPC Logging methods


hanzei:
Instead of using `mlog` all log calls should use the RPC methods `LogX`, e.g. [`LogDebug`](https://developers.mattermost.com/extend/plugins/server/reference/#API.LogDebug). This allows the server to gradually select which log messages to show in the server logs.

mo2menelzeiny:
@hanzei, I'm giving this a shot.
I don't know if this is the right place, but I'm having trouble with the testing.
Does any test cases cover the thrown errors? or do I have to manually test this?

My thoughts are for example
```
if err != nil {
        mlog.Error(fmt.Sprintf("Error creating autolinker: %+v: %v", c.Links[i], err))
}
```
I'm going to replace it with 
```
if err != nil {
	p.API.LogError(fmt.Sprintf("Error creating autolinker: %+v: %v", c.Links[i], err))
}
```
But how can I validate my change?

hanzei:
@mo2menelzeiny Please see my response in your PR: https://github.com/mattermost/mattermost-plugin-autolink/pull/102#issuecomment-602677926

PR #102, opened 2020-03-23
Migrate logging to use RPC Logging methods

mo2menelzeiny:
#### Summary
Migrates logging to use API Logging 

#### Ticket Link
Closes https://github.com/mattermost/mattermost-plugin-autolink/issues/99

### DIFF
--- a/server/autolinkplugin/config_test.go
+++ b/server/autolinkplugin/config_test.go
@@ -10,51 +10,54 @@ import (
 	"testing"
 )
 
-func TestOnConfigurationChangeInvalidConfig(t *testing.T) {
-	api := &plugintest.API{}
-	api.On("LoadPluginConfiguration",
-		mock.AnythingOfType("*autolinkplugin.Config")).Return(func(dest interface{}) error {
-		return errors.New("LoadPluginConfiguration Error")
+func TestOnConfigurationChange(t *testing.T) {
+
+	t.Run("Invalid Configuration", func(t *testing.T) {
+		api := &plugintest.API{}
+		api.On("LoadPluginConfiguration",
+			mock.AnythingOfType("*autolinkplugin.Config")).Return(func(dest interface{}) error {
+			return errors.New("LoadPluginConfiguration Error")
+		})
+
+		p := Plugin{}
+		p.SetAPI(api)
+
+		err := p.OnConfigurationChange()
+		assert.Error(t, err)
 	})
 
-	p := Plugin{}
-	p.SetAPI(api)
-
-	err := p.OnConfigurationChange()
-	assert.Error(t, err)
-}
-
-func TestOnConfigurationChangeInvalidLink(t *testing.T) {
-	conf := Config{
-		Links: []autolink.Autolink{
-			autolink.Autolink{
-				Name:     "existing",
-				Pattern:  ")",
-				Template: "otherthing",
+	t.Run("Invalid Autolink", func(t *testing.T) {
+		conf := Config{
+			Links: []autolink.Autolink{
+				autolink.Autolink{
+					Name:     "existing",
+					Pattern:  ")",
+					Template: "otherthing",
+				},
 			},
-		},
-	}
+		}
 
-	api := &plugintest.API{}
-	api.On("LoadPluginConfiguration",
-		mock.AnythingOfType("*autolinkplugin.Config")).Return(func(dest interface{}) error {
-		*dest.(*Config) = conf
-		return nil
+		api := &plugintest.API{}
+		api.On("LoadPluginConfiguration",
+			mock.AnythingOfType("*autolinkplugin.Config")).Return(func(dest interface{}) error {
+			*dest.(*Config) = conf
+			return nil
+		})
+
+		api.On("LogError",
+			mock.AnythingOfType("string"),
+			mock.AnythingOfType("string"),
+			mock.AnythingOfType("Autolink"),
+			mock.AnythingOfType("string"),
+			mock.AnythingOfType("string")).Return(nil)
+
+		api.On("UnregisterCommand", mock.AnythingOfType("string"),
+			mock.AnythingOfType("string")).Return((*model.AppError)(nil))
+
+		p := New()
+		p.SetAPI(api)
+		p.OnConfigurationChange()
+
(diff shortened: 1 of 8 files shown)


server/autolinkplugin/plugin.go
```diff
@@ -103,14 +103,14 @@ func (p *Plugin) ProcessPost(c *plugin.Context, post *model.Post) (*model.Post,
 
 			channel, cErr := p.API.GetChannel(post.ChannelId)
 			if cErr != nil {
-				mlog.Error(cErr.Error())
+				p.API.LogError(cErr.Error())
```
hanzei: Would you please give these log Messgesanges a bit more context. Something like `Failed to get channel` using fielded logging.

hanzei:
@mo2menelzeiny Regarding your question on https://github.com/mattermost/mattermost-plugin-autolink/issues/99#issuecomment-602293265: No there are no test that cover these lines of code. FYI: You can see which lines are covered by running `make coverage`. If you want to add unit test for them, it would be highly appreciated, but this is no requirement for merging this PR.

Manually testing is quite hard, as the failure cases are hard to produce.

mo2menelzeiny:
@hanzei awesome, I will work on the requested refinements

In regards to the test coverage, I noticed that all the cases where there's a check on an error object are not covered in the unit tests, is this intentional?

Also I started writing some tests for `config.go` however I have some issues with some coupled dependencies that I'm not able to mock, like for example this line
https://github.com/mattermost/mattermost-plugin-autolink/blob/37ebe2af40d3a79042f94ae23af67ae0df4c27dc/server/autolinkplugin/config.go#L28
is there an easy way to mock the `Compile()` funciton? because it's depending on `Config` which is declared in the same scope not passed.
https://github.com/mattermost/mattermost-plugin-autolink/blob/37ebe2af40d3a79042f94ae23af67ae0df4c27dc/server/autolinkplugin/config.go#L21
any ideas on that matter?

hanzei:
> In regards to the test coverage, I noticed that all the cases where there's a check on an error object are not covered in the unit tests, is this intentional?

No that is not intentional, this are just missing tests :wink:

hanzei:
Mocking is one way of testing, another would be to just test the error case. In this case an invalid `pattern` would cause an error:
https://github.com/mattermost/mattermost-plugin-autolink/blob/37ebe2af40d3a79042f94ae23af67ae0df4c27dc/server/autolink/autolink.go#L81-L83

server/autolinkplugin/plugin.go
```diff
@@ -120,14 +120,14 @@ func (p *Plugin) ProcessPost(c *plugin.Context, post *model.Post) (*model.Post,
 
 			channel, cErr := p.API.GetChannel(post.ChannelId)
 			if cErr != nil {
-				mlog.Error(cErr.Error())
+				p.API.LogError("Failed to get Channel", cErr.Error())
```
hanzei: `LogError` always expects an unequal number of arguments and will panic otherwise. Hence this has to be:
```suggestion
				p.API.LogError("Failed to get Channel", "error", cErr.Error())
```

server/autolinkplugin/plugin.go
```diff
@@ -120,14 +120,14 @@ func (p *Plugin) ProcessPost(c *plugin.Context, post *model.Post) (*model.Post,
...
 				return false
 			}
 			teamName := ""
 			if channel.TeamId != "" {
 				team, tErr := p.API.GetTeam(channel.TeamId)
 				if tErr != nil {
-					mlog.Error(tErr.Error())
+					p.API.LogError("Failed to get Team", tErr.Error())
```
hanzei: Same here
```suggestion
					p.API.LogError("Failed to get Team", "error", tErr.Error())
```

server/autolinkplugin/config_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"errors"
+	"github.com/mattermost/mattermost-plugin-autolink/server/autolink"
+	"github.com/mattermost/mattermost-server/v5/model"
+	"github.com/mattermost/mattermost-server/v5/plugin/plugintest"
+	"github.com/mattermost/mattermost-server/v5/plugin/plugintest/mock"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestOnConfigurationChangeInvalidConfig(t *testing.T) {
```
hanzei: Thanks for add a test :+1: 

Nit pick: If you have similar tests like `TestOnConfigurationChangeInvalidConfig` and `TestOnConfigurationChangeInvalidLink` you can group them by using [Subtests](https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks). This is more of can FYI then a request for change.

server/autolinkplugin/config_test.go
```diff
@@ -0,0 +1,63 @@
...
+		api.On("LoadPluginConfiguration",
+			mock.AnythingOfType("*autolinkplugin.Config")).Return(func(dest interface{}) error {
+			*dest.(*Config) = conf
+			return nil
+		})
+
+		api.On("LogError",
+			mock.AnythingOfType("string"),
+			mock.AnythingOfType("string"),
+			mock.AnythingOfType("Autolink"),
```
levb: Is this effectively the unique signature of the log call? (3rd parameter type Autolink?)

levb:
@DHaussermann this can use a general smoke test (and maybe checking that at least some errors are actually logged?) - perhaps bundled in with the release test?

server/autolinkplugin/config_test.go
mo2menelzeiny: based on this line
https://github.com/mattermost/mattermost-plugin-autolink/blob/433c563bfbb7039835838481d719aa791afa8dc7/server/autolinkplugin/config.go#L17
and this 
https://github.com/mattermost/mattermost-plugin-autolink/blob/433c563bfbb7039835838481d719aa791afa8dc7/server/autolinkplugin/config.go#L35

It should probably be
```suggestion
			mock.AnythingOfType("[]autolink.Autolink"),
```

mo2menelzeiny: right?

mo2menelzeiny:
awesome job guys, thank you!
@hanzei, @levb, @DHaussermann

Issue #87, opened 2020-01-31
Performance: do not fetch extra post data unless needed


levb:
Presently, processing each post autolink goes back to the server to get the Channel and Team names for it, to match the scope. https://github.com/mattermost/mattermost-plugin-autolink/blob/0c286d8baa27c03834c2240c0e20fd518b4ec83d/server/autolinkplugin/plugin.go#L104-L117

1. These should be requested lazily so if no links have `Scope` defined, there's no perf impact.
2. Long term we should consider a new hook `MessageWillBePostedEx` that would contain extended metadata (channel and team names in this case) and will not require extra roundtrips. cc @lieut-data 
3. Another long-term performance improvement here would be limiting the `Scope` to perform only exact matches on teams and channels, rather than the present `contains` (by keyword). Maybe we add new fields `Channel` and `Team` that are exact matches? This would allow pre-resolving all references on Config and comparing by ID, avoiding the extra per-message RPCs.

lieut-data:
@ali-farooq0, this will be an interesting case to think through with respect to autolink performance efforts. Knowing the RPC overhead, in particular, will influence where in the pipeline we should optimize.

mo2menelzeiny:
Hello @levb @hanzei 
I would love to work on this

PR #105, opened 2020-04-07
Performance: do not fetch extra post data unless needed

mo2menelzeiny:
#### Summary

1. Lazily resolves the scope channel and team only when the scope is defined
2. Adds gating conditions rather than nested if 
3. Moves the API call logic to a separate function
4. Increases coverage on `ProcessPosts`

#### Ticket Link

Closes https://github.com/mattermost/mattermost-plugin-autolink/issues/87


server/autolinkplugin/plugin.go
```diff
@@ -130,25 +150,18 @@ func (p *Plugin) ProcessPost(c *plugin.Context, post *model.Post) (*model.Post,
...
-			}
-
 			for _, l := range conf.Links {
 				if len(l.Scope) == 0 {
 					newText = l.Replace(newText)
-				} else if teamName != "" && contains(teamName, channel.Name, l.Scope) {
+					continue
+				}
+
+				teamName, channelName, rsErr := p.resolveScope(post.ChannelId)
```
levb, line 159: 1/5 suggestion: replace `resolveScope(channelId string) (string, string, *model.AppError)` with `inScope(scope, channelId string) bool`; then the loop becomes even simpler:
```go
	for _, l := range conf.Links {
		if p.inScope(l.Scope, post.ChannelId) {
			newText = l.Replace(newText)
		}
	}
```

mo2menelzeiny: Sounds logical but, I’m suggesting that it doesn’t have to be the resolve function.
4/5 a new `inScope` function is created that calls resolve and does the rest of the logic

mo2menelzeiny:
> Thanks for the contribution @mo2menelzeiny.
> Let me understand the overall logic here. We traverse the markdown tree with `Inspect` which basically calls passed function for every node(let's say `N` nodes). In the function itself for each `link` where we have a scope we call the `inScope` method. Let's assume we have `M` links, this makes `O(N*M)` calls of the `inScope`(and `resolveScope` respectively).
> Return values of `team.Name` and `channel.Name` will be the same for every call. Am I missing something? Why not call `resolveScope` method just once before the `Inspect`?

I don't know how I missed this, you're definitely right.

server/autolinkplugin/plugin.go
```diff
@@ -125,6 +120,12 @@ func (p *Plugin) ProcessPost(c *plugin.Context, post *model.Post) (*model.Post,
 	conf := p.getConfig()
 	postText := post.Message
 	offset := 0
+
+	channelName, teamName, rsErr := p.resolveScope(post.ChannelId)
```
iomodo, line 124: What we could do here is only call `resolveScope` if there is at least one non empty scope in the `cond.Links`. Otherwise we do not need `channelName` and `teamName`.

server/autolinkplugin/plugin.go
```diff
@@ -169,7 +170,7 @@ func (p *Plugin) ProcessPost(c *plugin.Context, post *model.Post) (*model.Post,
 					continue
 				}
 
-				if p.inScope(l.Scope, post.ChannelId) {
+				if p.inScope(l.Scope, channelName, teamName) {
 					newText = l.Replace(newText)
```
iomodo, line 174: `newText = l.Replace(newText)` is duplicated, can we not have it? There is no need of two `if`-s in this loop.

mo2menelzeiny: I think there is, because you want to replace the text in two cases:
1. No scopes at all
2. There is a matched scope

we can merge the if conditions with an `||` statement if that's what you're suggesting?

iomodo: yes, you can merge it or let `p.inScope` decide if `newText` should be changed.

server/autolinkplugin/plugin.go
```diff
@@ -92,6 +124,27 @@ func (p *Plugin) ProcessPost(c *plugin.Context, post *model.Post) (*model.Post,
...
+	if hasOneOrMoreScopes {
+		cn, tn, rsErr := p.resolveScope(post.ChannelId)
+		channelName = cn
+		teamName = tn
+
+		if rsErr != nil {
+			p.API.LogError("Failed to resolve scope", "error", rsErr.Error())
+		}
+	}
+
```
iomodo, line 147: May be we could put this logic inside `p.resolveScope`? I think it will be clearer.

server/autolinkplugin/plugin.go
```diff
@@ -59,22 +58,55 @@ func (p *Plugin) IsAuthorizedAdmin(userId string) (bool, error) {
...
-		} else if len(channelTeamSplit) == 1 {
-			if strings.EqualFold(channelTeamSplit[0], team) {
-				return true
-			}
-		} else {
-			mlog.Error("error splitting channel & team combination.")
+func (p *Plugin) resolveScope(channelId string) (string, string, *model.AppError) {
+	channel, cErr := p.API.GetChannel(channelId)
+	if cErr != nil {
+		p.API.LogError("Failed to get Channel", "error", cErr.Error())
```
iomodo, line 64: we are logging twice, here and outside of the function call. I don't think this log is necessary. You can just wrap the error and return.

server/autolinkplugin/plugin.go
```diff
@@ -59,22 +58,55 @@ func (p *Plugin) IsAuthorizedAdmin(userId string) (bool, error) {
...
+		return "", "", cErr
+	}
+
+	if channel.TeamId == "" {
+		return channel.Name, "", nil
+	}
+
+	team, tErr := p.API.GetTeam(channel.TeamId)
+	if tErr != nil {
+		p.API.LogError("Failed to get Team", "error", tErr.Error())
```
iomodo, line 74: same here

server/autolinkplugin/plugin.go
mo2menelzeiny: I'm having a problem agreeing with you. This attempt might appear that it will make the logic clearer but in fact it's more misleading:

1. It's not the responsibility of the resolve function to check when to resolve or not
2. It's the responsibility of the `ProcessPost` function to decide when to resolve or not, thus the logic is in it's right place
3. If we hide this logic inside the resolve it will appear misleading as the user will expect that the function resolves right away as per it's name

what do yo think?

iomodo: Makes sense @mo2menelzeiny, I would have this logic in a different method anyways, but it's not a big deal and it's your call.

mo2menelzeiny:
Hey guys, @DHaussermann @levb @iomodo 
are there any troubles with this?

levb:
@mo2menelzeiny Apologies for the wait, but we are heavily backlogged in QA, and this requires QA verification before it's merged.

mo2menelzeiny:
@levb @iomodo @DHaussermann 
I think the pull request is now ready for merging, am I right?

PR #113, opened 2020-05-12
Add DisableInHyperlinks to support more flexible use cases

grubbins:
#### Summary
Adds a new property of links called `DisableInHyperlinks`, which makes it possible to properly support both link-insertion and link-shortening for a similar regex.

#### Details
Suppose we want to support the following patterns:
* `MM-12345` -> `[MM-12345](http://jira.server/MM-12345)`
* `http://jira.server/MM-12345` -> `[MM-12345](http://jira.server/MM-12345)`

If we do not enable `WordMatch` on the first pattern, then it fails in posts like `see MM-12345, or MM-23456`. Conversely if we do enable `WordMatch` on the first pattern, then it also gets inserted inside hyperlinks (completely breaking them).

Fundamentally, we never want the first pattern to be applied inside a hyperlink, so I added a flag to prevent that. Then, to deal with the fact that one pattern replacement may insert a new hyperlink, we also need to walk the markdown tree for each link pattern (not apply all link patterns inside one markdown tree traversal).

#### Possible alternative approaches

An alternative I quite like would be to add a `MatchHyperlink` flag instead. In that case, a `Link` with this flag is only allowed to match an *entire* hyperlink and replace it (in the expected use case, it would replace it with a formatted inline link). Conversely `Link`s without this flag set would not be allowed to match any part of an existing hyperlink. I think this is actually a more intuitive result - however, it would not preserve behaviour for existing plugin users, so some kind of migration support would be needed.

For interest: my original fork of this plugin went even further, replacing the `Link` general concept with two separate kinds of pattern spec, one which only inserts new hyperlinks (and by definition never matches inside a hyperlink) and one which only replaces hyperlinks with formatted inline links (and by definition never matches anything *except* a whole hyperlink). However I see from the test case coverage that this plugin, despite the name, is intended for any kind of text replacement. I wonder if it would be better to separate out link insertion (text -> link), link formatting (link -> inline link) and general text replacement more explicitly.

### DIFF
--- a/server/autolinkplugin/plugin.go
+++ b/server/autolinkplugin/plugin.go
@@ -120,7 +120,6 @@ func (p *Plugin) isBotUser(userID string) (bool, *model.AppError) {
 func (p *Plugin) ProcessPost(c *plugin.Context, post *model.Post) (*model.Post, string) {
 	conf := p.getConfig()
 	postText := post.Message
-	offset := 0
 
 	hasOneOrMoreScopes := false
 	for _, link := range conf.Links {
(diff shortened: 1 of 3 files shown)


server/autolinkplugin/plugin.go
```diff
@@ -142,62 +141,69 @@ func (p *Plugin) ProcessPost(c *plugin.Context, post *model.Post) (*model.Post,
...
-		// never descend into the text content of a link/image
-		case *markdown.InlineLink:
-			return false
-		case *markdown.InlineImage:
-			return false
-		case *markdown.ReferenceLink:
-			return false
-		case *markdown.ReferenceImage:
-			return false
+	for _, link := range conf.Links {
```
levb, line 144: 2/5 I am a little concerned about this change, it may need some benchmarking. The new version invoked markdown.Inspect in a loop, and it may be expensive. Is it directly related to the functional change? Can the functional change be implemented in a single `Inspect` pass?

levb:
Also, @grubbins I like the "2 matcher types" approach, want to think some more about it.

server/autolinkplugin/plugin.go
grubbins: Yes, I think this is fundamentally required for this plugin, because each replacement has the capacity to change the markdown structure - therefore if we proceed to look at later replacements, we can no longer apply the rules regarding which markdown elements they can touch.

If you restore the original pattern of inspecting once, the unit tests I have added will break because one replacement inserts a new inline link and a later replacement (which should ignore that inline link) gets erroneously applied because it is still treating it as text.

grubbins:
> Also, @grubbins I like the "2 matcher types" approach, want to think some more about it.

If each replacement pattern had a 'Domain' property (a list) with values 'Text', 'AutoLink', 'InlineLink', then patterns can explicitly state where they should be applied.

Maintaining existing behaviour is not easy because existing behaviour is:
* existing patterns implicitly have `Domain=Text,AutoLink`
* *except* that if an earlier pattern inserts a new inline link (or even image I guess), later links will then hit that as well - fixed by traversing markdown tree for each pattern

If we wanted to break from existing behaviour to adopt this new approach, may Isuggest something more bold: at the same time we could rename the `Link` structure and rename the plugin itself? Since obfuscating credit card numbers has nothing to do with "links", I think the plugin name should reflect the more general use case (or the plugin should be split up to handle links and general text replacements).

server/autolinkplugin/plugin.go
levb: Right. Understood. 1/5 We should benchmark it on a few examples. I wonder if we can re-use the parsed markdown tree between the no-change links.

levb:
> at the same time we could rename the Link structure and rename the plugin itself?

That's kind of what I was thinking, but it'd be a kind of a big deal - the current plugin is popular. cc @aaronrothschild @hanzei.

grubbins:
Indeed, it would be a bigger change. I could imagine a migration path like this:
* new plugin name appears alongside old
* new plugin auto-imports config from old, if present, when first installed; could treat all links as text-only, unless they start with a protocol like `http://` in which case set the domain to Autolink
* in the MM release notes, encourage people to test and try the new plugin and disable old
* remove old one later

grubbins:
If you're open to merging this as-is, though, it would allow me to get my team on to the public version of this plugin instead of my fork. A more substantial split could come later?

server/autolinkplugin/plugin.go
iomodo: I agree that we should try to re-use parsed markdown or detect the structure change. Otherwise performance will suffer.

grubbins:
Any thoughts on progressing with this @levb ?

jasonblais:
@levb @iomodo for above ^

levb:
@grubbins I posted https://github.com/grubbins/mattermost-plugin-autolink/pull/1 with a simple benchmark that shows a 30% degradation on a rather simplistic test.

3/5 we shouldn't merge this in its current reincarnation. I'd like to further understand your original ask in details (not 100% sure if the test was illustrative)

levb:
OT: This PR drove the point home, we can use an Autolink2, redesigned with all the lessons learned, not backwards compatible. Better integrated with the markdown parser - to avoid re-parsing on changes.

jfrerich:
@levb, @iomodo it sounds like the benchmarking results have halted the merge. Any further suggestions on how to proceed?

grubbins:
A 30% degradation sounds bad in relative terms, but I'm less sure what the actual user impact is. In my organisation (now nearing 1000 users) we have been running my fork of this plugin for a long time and never had any concern with performance. I would (4/5) value the functionality described above much more highly than a performance impact that we seem unable to notice, so we'll have to continue with the fork for now. I appreciate though that it's hard to get a sense for the real impact. Maybe this patch could run on the MM community instance for a while to get real-world user feedback?

The ideas about a new version of the plugin do sound good so I'd look forward to a future version!

lindy65:
Hi @grubbins :) I'm helping review stale PRs and wondered whether you're still working on this one and/or what is required to push it over the finish line? Let me know if it can be closed if no longer a priority?

//cc @levb @iomodo

grubbins:
As I understand it I think the PR is essentially rejected, mostly for performance concerns, and the discussions about possibly restructuring this plugin / splitting it up are incomplete.

Meanwhile, we're continuing to run my fork of the plugin at my workplace so we can do both link shortening and link insertion for JIRA tickets (which is a popular feature).

lindy65:
Thanks for letting me know @grubbins :)  I'll close this PR off for now then.

Issue #109, opened 2020-04-21
`/autolink list` enable searching patterns and templates


levb:
Currently, one can `/autolink list` search by link name, but not by the contents of `Pattern` and `Template`. Add the capability.

GrigalashviliT:
I want to work on this

PR #117, opened 2020-05-27
GH-109 '/autolink list' enable searching patterns and templates

GrigalashviliT:
#### Summary
Adds possibilities to list links according to Template or Pattern values 

#### Ticket Link
  Fixes https://github.com/mattermost/mattermost-plugin-autolink/issues/109

### DIFF
--- a/server/autolinkplugin/command.go
+++ b/server/autolinkplugin/command.go
@@ -12,7 +12,8 @@ import (
 	"github.com/mattermost/mattermost-plugin-autolink/server/autolink"
 )
 
-const template, pattern = "Template", "Pattern"
+const optName, optTemplate, optPattern, optScope, optDisabled, optDisableNonWordPrefix, optDisableNonWordSuffix, optWordMatch = "Name", "Template", "Pattern", "Scope", "Disabled", "DisableNonWordPrefix", "DisableNonWordSuffix", "WordMatch"
+const command = "/autolink"
 
 const helpText = "###### Mattermost Autolink Plugin Administration\n" +
 	"<linkref> is either the Name of a link, or its number in the `/autolink list` output. A partial Name can be specified, but some commands require it to be uniquely resolved.\n" +
@@ -80,7 +81,7 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, commandArgs *model.CommandArg
 	}
 
 	args := strings.Fields(commandArgs.Command)
-	if len(args) == 0 || args[0] != "/autolink" {
+	if len(args) == 0 || args[0] != command {
 		return responsef(helpText), nil
 	}
 
@@ -92,10 +93,10 @@ func executeList(p *Plugin, c *plugin.Context, header *model.CommandArgs, args .
 	var refs []int
 	var err error
 
-	if len(args) > 0 && (args[0] == template || args[0] == pattern) {
-		links, refs, err = parseLinkRefByTemplateOrPattern(p, header, args...)
+	if len(args) > 0 && (args[0] == optTemplate || args[0] == optPattern) {
+		links, refs, err = searchLinkRefByTemplateOrPattern(p, header, args...)
 	} else {
-		links, refs, err = parseLinkRef(p, false, args...)
+		links, refs, err = searchLinkRef(p, false, args...)
 	}
 	if err != nil {
 		return responsef("%v", err)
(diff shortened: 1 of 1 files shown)


server/autolinkplugin/command.go
```diff
@@ -12,13 +12,16 @@ import (
 	"github.com/mattermost/mattermost-plugin-autolink/server/autolink"
 )
 
+const template, pattern = "Template", "Pattern"
```
levb, line 15: 2/5 if we parameterize command names we should do it for all commands, and use more intention-revealing names, perhaps `optTemplate` etc.

@iomodo we should file a ticket to convert autolink to autocomplete!

server/autolinkplugin/command.go
```diff
@@ -85,7 +88,15 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, commandArgs *model.CommandArg
 }
 
 func executeList(p *Plugin, c *plugin.Context, header *model.CommandArgs, args ...string) *model.CommandResponse {
-	links, refs, err := parseLinkRef(p, false, args...)
+	var links []autolink.Autolink
+	var refs []int
+	var err error
+
+	if len(args) > 0 && (args[0] == template || args[0] == pattern) {
+		links, refs, err = parseLinkRefByTemplateOrPattern(p, header, args...)
```
levb, line 96: 0/5 rename the `parseLink...` functions to `searchLink...` to be more intention-revealing?

server/autolinkplugin/command.go
```diff
@@ -329,6 +340,31 @@ func parseLinkRef(p *Plugin, requireUnique bool, args ...string) ([]autolink.Aut
 	return links, found, nil
 }
 
+func parseLinkRefByTemplateOrPattern(p *Plugin, header *model.CommandArgs, args ...string) ([]autolink.Autolink, []int, error) {
+	links := p.getConfig().Sorted().Links
+	if len(args) == 1 {
+		return links, nil, nil
+	}
+
+	restOfCommand := header.Command[10:]
```
levb, line 349: 1/5 `10` must be a length of something... better expressed as len(x)

server/autolinkplugin/command.go
```diff
@@ -12,7 +12,8 @@ import (
 	"github.com/mattermost/mattermost-plugin-autolink/server/autolink"
 )
 
-const template, pattern = "Template", "Pattern"
+const optName, optTemplate, optPattern, optScope, optDisabled, optDisableNonWordPrefix, optDisableNonWordSuffix, optWordMatch = "Name", "Template", "Pattern", "Scope", "Disabled", "DisableNonWordPrefix", "DisableNonWordSuffix", "WordMatch"
```
levb, line 15: 1/5 format as 1/line for readability/typography

server/autolinkplugin/command.go
```diff
@@ -12,7 +12,8 @@ import (
...
+const command = "/autolink"
```
levb, line 16: 0/5 make it a bit more specific, intention-revealing; maybe `autolinkCommand` or `cmdAutolink`?

server/autolinkplugin/command.go
iomodo: It already has been filed https://github.com/mattermost/mattermost-plugin-autolink/issues/118

server/autolinkplugin/command.go
```diff
@@ -329,6 +350,31 @@ func parseLinkRef(p *Plugin, requireUnique bool, args ...string) ([]autolink.Aut
...
+	if len(args) == 1 {
+		return links, nil, nil
+	}
+
+	restOfCommand := header.Command[len(autolinkCommand):]
+	restOfCommand = restOfCommand[strings.Index(restOfCommand, args[0])+len(args[0]):]
+	value := strings.TrimSpace(restOfCommand)
+
+	found := []int{}
+	for i, l := range links {
```
jfrerich, line 364: Outside the scope of this PR, but after just a quick look, I don't see why we need to return the array of indexes. Couldn't the caller just iterate over the returned links? @levb

levb: I can't remember off the top of my head, but does it have to do with the indices listed with `/autolink list`?

DHaussermann:
@GrigalashviliT thank you for this improvement! Sorry for the delay in testing this. I've taken a look and it seems to be working as expected.

I have only one minor point to bring up. Would it take much effort to make the `value` case  insensitive. For example `/autolink list Template  scrum` would work the same as `/autolink list Template  SCRUM`.  Since something like Jira issue keys would have upper case letters. So, if it's simple to make the value case insensitive, it would improve the usability even more.

DHaussermann:
Thanks @GrigalashviliT for this enhancement!

I see there is still a `Docs/Needed` label on this. The help command has been updated. Should we also update the table in the read-me with the new list command?
@levb or @iomodo, I'm not certain who does read-me updates or who removes this label.

PR #123, opened 2020-06-20
Fix `/autolink test` command for disabled links

### LINKED ISSUE #119: /autolink test seems to ignore disabled links
From what I can see (using 1.2.0 and mattermost 5.23.0), using /autolink test Foo some text just replies "Foo: no change" even when the text should match... Enablink the link and running the command again then gives the expected output.  But that defeats the whole purpos of disabling a link for testing...

levb:
Fixes #119 

Also adds a "test library" of use-case specific links

### DIFF
--- a/server/autolink/lib_jira_test.go
+++ b/server/autolink/lib_jira_test.go
@@ -1,0 +1,125 @@
+package autolink_test
+
+import (
+	"testing"
+
+	"github.com/mattermost/mattermost-plugin-autolink/server/autolink"
+)
+
+var jiraTests = []linkTest{
+	{
+		"Jira example",
+		autolink.Autolink{
+			Pattern:  "(MM)(-)(?P<jira_id>\\d+)",
+			Template: "[MM-$jira_id](https://mattermost.atlassian.net/browse/MM-$jira_id)",
+		},
+		"Welcome MM-12345 should link!",
+		"Welcome [MM-12345](https://mattermost.atlassian.net/browse/MM-12345) should link!",
+	}, {
+		"Jira example 2 (within a ())",
+		autolink.Autolink{
+			Pattern:  "(MM)(-)(?P<jira_id>\\d+)",
+			Template: "[MM-$jira_id](https://mattermost.atlassian.net/browse/MM-$jira_id)",
+		},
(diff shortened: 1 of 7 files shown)


levb:
BTW, @hanzei when I run `make all` locally I get lint errors, do I need to do something in my local lint config to fix it and if so should it be checked in?
image

server/autolink/lib_jira_test.go
```diff
@@ -0,0 +1,125 @@
+package autolink_test
+
+import (
+	"testing"
+
+	"github.com/mattermost/mattermost-plugin-autolink/server/autolink"
+)
+
+var jiraTests = []linkTest{
```
hanzei, line 9: I don't see the need to put this into a global variable. No other tests needs it.

server/autolinkclient/client.go
```diff
@@ -40,7 +40,7 @@ func NewClientPlugin(api PluginAPI) *Client {
 
 func (c *Client) Add(links ...autolink.Autolink) error {
 	for _, link := range links {
-		linkBytes, err := json.Marshal(&link)
+		linkBytes, err := json.Marshal(link)
```
hanzei, line 43: Good catch

server/autolink/lib_jira_test.go
levb: I just want to move the relevant template definitions to the top of the file, as de-facto documentation. Eventually - should be exposed as a bunch of godocs? (Orthogonal to any other efforts to "templatize" autolink deployments)

server/autolinkclient/client.go
levb: It's a lint-only issue, I think - same JSON, but was an annoyance.

levb:
@aaronrothschild this was FYI  as the beginnings of a template lib.

server/autolink/lib_jira_test.go
hanzei: I just like to have everything related to a test bundled into the `TestX` method. But at the end, that is personal preference.

> Eventually - should be exposed as a bunch of godocs?

What is the value in having godocs for tests?

Issue #137, opened 2020-10-16
Enable slash commands by default


hanzei:
_Issue created from a [Mattermost message](https://community-release.mattermost.com/_redirect/pl/o57x5rw977gp5rj7oq1n7mx7sh) *by @aaronrothschild*._

PR #139, opened 2020-10-22
Enable slash command by default

hanzei:
#### Summary
Given that in Cloud the slash command are the only way to configure the plugin, we should enabled it by default. The feature has soaked for long enough. 

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-autolink/issues/137

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -21,7 +21,7 @@
         "key": "EnableAdminCommand",
         "display_name": "Enable administration with /autolink command",
         "type": "bool",
-        "default": false
+        "default": true
       },
       {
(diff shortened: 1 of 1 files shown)


levb (approved):

aaronrothschild (approved):
LGTM

jfrerich (approved):

DHaussermann (approved):
Tested and passed. 
- Tested Fresh install on multiple servers
- When enable radio is True, slash command radio defaults to true as well. 
- No issues enabling or disabling

LGTM!

Issue #136, opened 2020-10-12
Failing tests do not appear properly in GitHub/CI


levb:
@levb @iomodo So it seems that I missed the fact that my change actually makes already existing tests fail. I ignored the failed part of CI on GitHub because of its name - "coverage". I assumed it's about percentage of coverage and not about tests. I've logged into Circle CI a couple of minutes ago to see the details.

I think that unit tests treating the `(` character as non-boundary need a discussion.

But the second group of fails is much more interesting: more than one pattern/template seems to be applied to a single text fragment and MM-12345 gets expanded twice.

I'm going to experiment a bit with the code despite my lack of knowledge about Go ;)

BTW, with a more complicated pattern `(?P<MattermostNonWordPrefix>(^|^\(|\s|\s\())` tests pass but I have to think if it would be a solution or just a dirty workaround.

_Originally posted by @wodny in https://github.com/mattermost/mattermost-plugin-autolink/pull/130#issuecomment-700146504_

catalintomai:
Fix for autolink only (for now).

PR #140, opened 2020-10-22
Update CI

hanzei:
#### Summary
- Drop plugin-ci/deploy-release
- Add test job: `coverage` does require `test` because if `test` fails `coverage` will surely also fail
- Add nighly job

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-autolink/issues/136

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,6 +5,18 @@ orbs:
 
 workflows:
   version: 2
+  nightly:
+    triggers:
+      - schedule:
+          cron: "0 0 * * *"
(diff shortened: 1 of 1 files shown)


levb (approved):
why nightly? Best practice?

jfrerich (approved):

hanzei:
> why nightly? Best practice?

We are using the latest version of the Plugin CI orb here which means updates of that orb might cause CI for this to fail. A nightly will spot such failures before we notice them in a PR.

Issue #138, opened 2020-10-16
slash-commands not looking for exact match


TheLure:
Verified in 1.2.1: The /autolink commands does not look fot the exact link name when you try to manipulate links. This makes it impossible to manipulate some links.

Example:

/autolink add testA
This is fine.
/autolink add testA2
Also fine.
/autolink set testA Pattern 123456789
returns:
"testA" matched more than one link: ["testA" "testA2"]

Correct behaviour would be to only look for a exact match.

jrepe:
hey, can I grab this one?

p3k:
any update on this?

p3k:
just out of sheer desperation i tried to use the index numbers of each autolink to change my version of the `testA` link – and what can i say, this works!

```plain
/autolink list
1: testA
Pattern: ``
Template: ``
2: testA2
Pattern: ``
Template: ``

/autolink set 1 Pattern 123456789
1: testA
Pattern: 123456789
Template: ``
2: testA2
Pattern: ``
Template: ``
```

PR #142, opened 2020-10-24
Slash-commands now look for exact match

jrepe:
#### Summary

The pull request fixes the issue where `/autolink [command]` doesn't look for exact match when manipulating links. 

However, the current implementation might be too naive, because it'll affect all commands and not just those that manipulate links, so I'd appreciate feedback if that is desirable or not, otherwise I'll need to give more thought to how to fix this only for specific commands.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-plugin-autolink/issues/138

### DIFF
--- a/server/autolinkplugin/command.go
+++ b/server/autolinkplugin/command.go
@@ -336,7 +336,7 @@ func searchLinkRef(p *Plugin, requireUnique bool, args ...string) ([]autolink.Au
 
 	found := []int{}
 	for i, l := range links {
-		if strings.Contains(l.Name, args[0]) {
+		if strings.Compare(l.Name, args[0]) == 0 {
 			found = append(found, i)
 		}
 	}


jfrerich (approved):
Nicely done :) Would you mind adding a test for this particular code change?

jrepe:
Of course, will do 👍

levb (approved):
I will kinda miss the regex there, but I agree that it was probably more confusing in partial match situations.

LGTM

jasonblais:
@DHaussermann kind reminder to help with reviews, thank you!

DHaussermann:
Hey @jrepe thank you so much for fixing this issue! This bug was especially annoying. 

I've done some testing and the issue is fixed but, I did find a small issue introduced by this solution. In cases where you're using `list` to do a search, the fact that it works as a begins with is actually useful. For example if I have `Jira-ABC`. `Jira-123` and `Jira-EXT` I can type `/autolink list Jira` to show all 3 records

With this change, it will not pull them back as list must now be an exact match. I have tested this with all of the commands and it solves the issue for manipulating the data. But, would it be possible to make an exception for `list` to preserve the previous behavior?

jrepe:
Hey, thanks for the comment, I'll look into it. I imagined that this might have some consequences somewhere else that would be undesirable. Sorry for the delay though, I've had little time lately, but will get on this as soon as I can.

jfrerich:
How's it going, @jrepe? Sounds like you have been busy, but I just wanted to check in to see if you have any further questions or need any assistance?

lindy65:
Hi @jrepe - I'm helping review stale PRs and, as we haven't heard from you in 6 months, I'll close this PR for now. Please feel free to re-open it at any time if you're able to prioritize it.

I've re-added the "Up for Grabs" label to https://github.com/mattermost/mattermost-plugin-autolink/issues/138

//cc @jfrerich @levb

PR #143, opened 2020-10-25
[GH-118] Add slash command autocomplete functionality

### LINKED ISSUE #118: Add slash command autocomplete functionality
From 5.24 MM server has a slash command autocomplete functionality feature. An API can be found [here](https://github.com/mattermost/mattermost-server/blob/master/model/command_autocomplete.go#L28)
To implement the feature add `AutocompleteData` struct to the `Command` when registering a slash command.

Similar PRs:
For Todo plugin https://github.com/mattermost/mattermost-plugin-todo/pull/69
For Agenda plugin https://github.com/mattermost/mattermost-plugin-agenda/pull/40

ByeongsuPark:
#### Summary

This PR adds slash command autocomplete functionality.

#### Ticket Link

Fixes #118


levb:
@ByeongsuPark, thank you for the contribution! Do you think you can also add dynamic select support, as appropriate, to select the links to view/edit? It has been one of the biggest usability issues with the command, especially for links that do not have names. If you prefer, it can be done as a separate PR?

ByeongsuPark:
Hi, @levb. I can't sure whether I can implement the feature you explained for now. How about open a new issue about the dynamic select support feature so that someone else also can work on the issue?

jfrerich:
@levb, would you mind creating the new GitHub issuea and reviewing this PR as is?

levb (approved):

iomodo (approved):
LGTM! 👍

DHaussermann:
@ByeongsuPark Thanks for this enhancement. This makes for a huge usability improvement. 

I have now done a round of testing and this is mostly working as expected. I see that the `test` command was left out which to me, seems like it may have been intentional. But, there are a couple places where the auto-complete functionality is incomplete. 

1. When using `set`, `Scope` should be one of the available options. ex `/autolink set jira Scope MyTeam`
2. When using `list` it is also possible to get results based on the value of `Template` and `Pattern` so, in addtion to `[name]` these should appear as optional for commands such as `/autolink list Pattern mmtest` or `/autolink list Template ${jira_id}` 

This PR is a great addition as is, but let me know if your able to add these 2 points as well.

ByeongsuPark:
@DHaussermann  Yes, I can add it. I didn't commit it yet but Should the result look like the images below?

스크린샷, 2020-12-15 16-57-41
스크린샷, 2020-12-15 16-54-26

DHaussermann:
@ByeongsuPark Sorry for the late response. These look good. 
One small thing is missing for consistency with this and other plugins. In the 1st screen - The options for `Template` and `Scope` should also include the `(optional)` beside them the same way `[name]` already has. 

Thanks for taking the time to further improve this. Much appreciated.

ByeongsuPark:
@DHaussermann I fixed and commit it. Now it looks like the image below.
Also, thanks to your comment, I registered the missed test autocomplete command data
스크린샷, 2020-12-28 17-26-59

DHaussermann:
Thanks @ByeongsuPark but, I do see one last minor issue in the screenshot. `Scope` and `Pattern` do not take the `[]` . This should be unchanged from before as these are options you must select. `name` does have the `[]` because this is the value that will be used if you start to type in some text without selecting anything.

Sorry if my last comment was not completely clear. The UX for Autolink is a bit different from other plugins so I'm curious if it's trivial for you to make this change?

ByeongsuPark:
@DHaussermann I can make it change. But I have a question. 
For autocomplete command for `list`,  
Is `Template` what you want to mention instead of `Scope`?
I'm confused because you said that `Scope` does not take the `[]`.

DHaussermann:
@ByeongsuPark yes, exactly. Sorry I made an error when adding the comment :)

`Template` and `Pattern` are what must be changed. I have now edited my comment to correct it. 

_Side note for clarity_: `Scope` is not available when using `list` and on your last commit `Scope` looks correct when using `set` so no changes are needed for Scope

ByeongsuPark:
@DHaussermann 
I've committed a fix.
For autocomplete `list` command, `Pattern` and `Scope` now does not take `[]`

DHaussermann (approved):
Tested and passed
- Reviewed all commands including changes to `list` subcommand
- Autocomplete now matches convention in all cases
- Tested in Mobile client as well as a precaution
- No issues found 

LGTM!

@ByeongsuPark thank you so much for your continued efforts on this to implement the remaining changes. Your contribution is very much appreciated!

PR #153, opened 2021-03-27
Formatting fix for command

bbodenmiller:


hanzei (changes requested):
Good catch! Would you be open on also updating the Usage test in the following table?

levb (approved):

bbodenmiller:
@hanzei updated

hanzei (approved):
Awesome :tada:

PR #155, opened 2021-04-15
Fixed setting Mattermost-Plugin-ID header

levb:
#### Summary
`Mattermost-Plugin-ID` header was `Add`ed, not `Set` in the request headers, resulting in a vulnerability.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-34892

- Note: @srkgupta @crspeller is this vulnerability present in other plugins? Also, 1/5 we should probably handle `Mattermost-Plugin-ID` on the server since most plugins use `mux` and the handlers don't have access to `plugin.Context` easily.
- Note: @catalintomai @iomodo @aaronrothschild we'll need a new release for this.


srkgupta (approved):
Thanks @levb. Went ahead and tested the vulnerability, even before the other Dev reviews and the issue is fixed now and is working fine. Please feel free to re-request review if there are any additional changes based on Dev review. Thanks.

catalintomai (approved):

PR #176, opened 2021-11-22
dist: build linux/darwin arm64 targets

cpanato:
#### Summary
Adding the linux/darwin arm64 targets, we start building arm64 for the server and will start adding the targets for the plugins for the ones we prepackage in the build

related to: https://github.com/mattermost/mattermost-server/pull/18491

#### Ticket Link
related to https://mattermost.atlassian.net/browse/DOPS-633

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -7,11 +7,13 @@
   "release_notes_url": "https://github.com/mattermost/mattermost-plugin-autolink/releases/tag/v1.2.1",
   "icon_path": "assets/icon.svg",
   "version": "1.2.1",
-  "min_server_version": "5.16.0",
+  "min_server_version": "5.37.0",
   "server": {
     "executables": {
       "linux-amd64": "server/dist/plugin-linux-amd64",
+      "linux-arm64": "server/dist/plugin-linux-arm64",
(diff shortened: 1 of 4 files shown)


plugin.json
```diff
@@ -7,11 +7,13 @@
   "release_notes_url": "https://github.com/mattermost/mattermost-plugin-autolink/releases/tag/v1.2.1",
   "icon_path": "assets/icon.svg",
   "version": "1.2.1",
-  "min_server_version": "5.16.0",
+  "min_server_version": "5.37.0",
```
levb, line 10: @hanzei Is this actually correct, or should we require 6.0?

levb (approved):
LGTM, but I am starting to get even more concerned about the plugin bundle size; wonder if we should switch to platform-specific bundles soon.

plugin.json
cpanato: following the same approach for the other PRs that Ben suggest to update this. i can make a change to what fits best

hanzei: 5.37.0 is correct.

hanzei (approved):

Issue #181, opened 2022-01-13
Autolink admin user IDs requires spaces between commas


bbodenmiller:
![image](https://user-images.githubusercontent.com/1192780/149370550-ca286e57-cf79-4d91-aa9d-13051bc3ab32.png)

Appears to require spaces between commas are the admin features do not work.

hanzei:
@dipak-demansol Could you please try to reproduce this?

DHaussermann:
After further exploration - it seems that saving a 2nd user (or additional users) to the config only work intermittently.  

Regardless of if there is a whitespace character - saving the string with multiple users only works some of the time. This happpens regadless of if there is whitespace or if the string is valid such as `xntwk983abgbdgixdfhha3xcsc,x8xo6zmhhfnc3p4w6eqofky9uw`

We have confirmed that when the issue reproduces, the value also looks correct in the configuration file
![image](https://user-images.githubusercontent.com/35588783/151821339-fa01343c-7621-496e-b1e3-0ff97e54642c.png)

Removing the value and resaving the **same** string can solve the issue. It is not known why this is intermittent.

cc @hanzei

hanzei:
Might be related to https://github.com/mattermost/mattermost-plugin-autolink/issues/180

PR #184, opened 2022-02-23
Fix Autolink admin config parsing

hanzei:
#### Summary
The plugin was falsely storing it's config in UperCase names, which lead to the webapp not being able to rea

This PR make use of `json.Marshal` to convert the plugin config to an `map[string]Interface{}` and hence json tags are used everywhere now.

There is one migration left. If an admin (after updating to this PR) sees the empty list in the system console and inputs something, the existing admin list will be overwritten. Running `/autolink add` or `/autolink delete` brings the configuration in the right state again. 

#### Ticket Link
https://github.com/mattermost/mattermost-plugin-autolink/issues/180
https://github.com/mattermost/mattermost-plugin-autolink/issues/181

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -21,19 +21,19 @@
     "header": "Configure this plugin directly in the `config.json` file, or using the `/autolink` command. Learn more [in our documentation](https://github.com/mattermost/mattermost-plugin-autolink/blob/master/README.md).\n\n To report an issue, make a suggestion, or contribute, [check the plugin repository](https://github.com/mattermost/mattermost-plugin-autolink).",
     "settings": [
       {
-        "key": "EnableAdminCommand",
+        "key": "enableadmincommand",
         "display_name": "Enable administration with /autolink command:",
         "type": "bool",
         "default": true
       },
       {
-        "key": "EnableOnUpdate",
+        "key": "enableonupdate",
         "display_name": "Apply plugin to updated posts as well as new posts:",
         "type": "bool",
         "default": false
       },
       {
-        "key": "PluginAdmins",
+        "key": "pluginadmins",
         "display_name": "Admin User IDs:",
         "type": "text",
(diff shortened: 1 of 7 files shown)


levb (approved):

mickmister (approved):

catalintomai (approved):

dipak-demansol:
> @dipak-demansol gentle reminder to review this PR

@hanzei @DHaussermann  i'm not able to regenerate the 180 & 181 issue with the master Build. Its working fine with the master build.

hanzei:
@dipak-demansol Did you test with a clean plugin config i.e. `PluginSettings.Plugins.mattermost-autolink` is empty?

dipak-demansol:
> @dipak-demansol Did you test with a clean plugin config i.e. `PluginSettings.Plugins.mattermost-autolink` is empty?

Yes

hanzei:
@dipak-demansol Could you please test with the following steps:

1. Set an autolink admin via the system console
2. Run `/autolink add foo` to create an link
3. Check the list of autolink admins in the system console

Observed behaviour on `master`:
The list is empty

Expected behaviour:
The list contains the admins set in step 1

dipak-demansol (approved):
@hanzei tested as you told and its LGTM, adding the cases with that i tested, pls let me know if i miss any case to test.

1) adding two member-Users id in console,those 2 user is able to use autolink slash command.
2) after performing First case, in the admin system console i refresh the page and still all data is there and those Two users are able to use to autolink slash command. (its a case that ben told me to check) (issue 181)
3) by adding space between the two id, expected result is slash command should not work for both user.:- as expected.
4) by adding space before the first id, still both user is able to use the slash command and i don't have any issue with that so it LGTM.
5) by adding space before the first id and adding space after the last id, both User are able to use the slash command.
6) added the user-id on the config file and restarted the server then i'm able to see that config value in the system-console. (issue 180)

Issue #188, opened 2022-03-21
Autolink logs "Markdown autolink did not match range text"


catalintomai:
Hi team, not sure if this is the correct place to get some feedback on this:
Is there a reason why entering www.whatever.com in the chat box the error below appears on the log?

```
{"timestamp":"2022-03-15 05:20:36.471 +01:00","level":"error","msg":"Markdown autolink did not match range text, '[http://www.whatever.com'](http://www.whatever.com%27/) != 'www.whatever.com'","caller":"app/plugin_api.go:928","plugin_id":"mattermost-autolink"}
However, this does not seems to affect when entering whatever.com or http://whatever.com/ on the chat room.
```
Is this expected?

Mattermost message: https://community.mattermost.com/core/pl/rpyzuujifif7tkna48b7jgcpor
Jira ticket: https://mattermost.atlassian.net/browse/MM-42669

dipak-demansol:
@catalintomai issue is regenerated, i will share a video on our developer Channel.

dipak-demansol:
i regenerated the issue, the log is similar but its not exactly same, i got this log 
`"level":"error","msg":"Markdown autolink did not match range text, 'http://www.whatever.com' != 'www.whatever.com'","caller":"app/plugin_api.go:940","plugin_id":"mattermost-autolink"}
`

maisnamrajusingh:
@sibasankarnayak @dipak-demansol  lets look into this one and see how often this occurs and if there is a fix needed.

hanzei:
@maisnamrajusingh Can you please provide a status update on this issue?

hanzei:
@maisnamrajusingh Gentle reminder

maisnamrajusingh:
@hanzei  this was already communicated earlier during one of the calls and siba was issuing a fix for it.

PR #201, opened 2022-11-05
[MM-42669] Avoid superfluous error messages on some autolinks

levb:
#### Summary
For historical reasons, the plugin was skipping autolinks that started with www and were not canonically formatted; it was also LogError-ing unnecessarily. 

Also improved the code typography.

#### Ticket Link
#188 
https://mattermost.atlassian.net/browse/MM-42669

### DIFF
--- a/server/autolinkplugin/plugin.go
+++ b/server/autolinkplugin/plugin.go
@@ -109,7 +109,9 @@ func (p *Plugin) inScope(scope []string, channelName string, teamName string) bo
 
 func (p *Plugin) ProcessPost(c *plugin.Context, post *model.Post) (*model.Post, string) {
 	conf := p.getConfig()
-	postText := post.Message
+
+	message := post.Message
+	changed := false
 	offset := 0
 
 	hasOneOrMoreScopes := false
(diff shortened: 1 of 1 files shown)


server/autolinkplugin/plugin.go
```diff
@@ -136,86 +138,83 @@ func (p *Plugin) ProcessPost(c *plugin.Context, post *model.Post) (*model.Post,
...
 	})
-	if post.Message != postText {
-		post.Message = postText
-	}
-
-	post.Hashtags, _ = model.ParseHashtags(post.Message)
 
+	if changed {
+		post.Message = message
+		post.Hashtags, _ = model.ParseHashtags(message)
```
jespino, line 216: This is now inside the if, looks like something that makes sense, because if I haven't change it, the original post should be having properly processes the hashtags already, but is a subtle change in the behavior and I just want to let you know to ensure that it is on purpose change.

jespino (approved):
LGTM, really nice improvement in the code itself.

server/autolinkplugin/plugin.go
levb: Exactly. I think it was basically a bug/redundant before.

hanzei (approved):
Very nice rewrite, LGTM :+1:

levb:
@DHaussermann 2/5 skip the QA on the PR, but do some release testing around "autolinks", i.e. URLs mixed into the text.

hanzei:
Let's merge this and test it in release testing

Issue #207, opened 2023-02-22
Introduce CRUD operations for Autolink Client


hanzei:
The Autolink Client API currently only supports an `Add` operations. A remove, update and delete operation should be added to serve use cases like the one in https://github.com/mattermost/mattermost-plugin-jira/pull/871#issuecomment-1415838868.

PR #213, opened 2023-04-10
[GH-207] Added CRUD operations in autolink plugin

Kshitij-Katiyar:
**Summary** 

- The Autolink Client API currently only supports Add operation only. So added CRUD operations in the Autolink plugin.

**Issue** 

- [Issue #207](https://github.com/mattermost/mattermost-plugin-autolink/issues/207)

### DIFF
--- a/server/autolinkclient/client_test.go
+++ b/server/autolinkclient/client_test.go
@@ -2,7 +2,7 @@ package autolinkclient
 
 import (
 	"errors"
-	"io/ioutil"
+	"io"
 	"net/http"
 	"strings"
 	"testing"
@@ -66,7 +66,7 @@ func TestDeleteAutolinks(t *testing.T) {
 		{
 			name: "delete the autolink",
 			setupAPI: func(api *plugintest.API) {
-				body := ioutil.NopCloser(strings.NewReader("{}"))
+				body := io.NopCloser(strings.NewReader("{}"))
 				api.On("PluginHTTP", mock.AnythingOfType("*http.Request")).Return(&http.Response{StatusCode: http.StatusOK, Body: body})
 			},
 		},
@@ -103,7 +103,7 @@ func TestGetAutolinks(t *testing.T) {
 		{
 			name: "get the autolink",
 			setupAPI: func(api *plugintest.API) {
-				body := ioutil.NopCloser(strings.NewReader("{}"))
+				body := io.NopCloser(strings.NewReader("{}"))
 				api.On("PluginHTTP", mock.AnythingOfType("*http.Request")).Return(&http.Response{StatusCode: http.StatusOK, Body: body})
 			},
 		},


Kshitij-Katiyar:
@m1lt0n Not able to add you as a reviewer here.

server/api/api.go
```diff
@@ -114,12 +118,85 @@ func (h *Handler) setLink(w http.ResponseWriter, r *http.Request) {
...
+		h.handleError(w, errors.New("autolink name or pattern should not be empty"))
+		return
+	}
+
+	links := h.store.GetLinks()
+	found := false
+	for i := 0; i < len(links); i++ {
+		if links[i].Name == autolinkName || links[i].Pattern == autolinkName {
+			links = append(links[:i], links[i+1:]...)
+			found = true
```
mgdelacroix, line 142: If we decide to match both `Name` and `Pattern`, we should append all links that are not a match, as slicing the links and adding them will include possible previous or next matches, and duplicate information

server/api/api.go
```diff
@@ -114,12 +118,85 @@ func (h *Handler) setLink(w http.ResponseWriter, r *http.Request) {
...
+	autolinkName := r.URL.Query().Get(autolinkclient.AutolinkNameQueryParam)
+	if autolinkName == "" {
+		h.handleError(w, errors.New("autolink name or pattern should not be empty"))
+		return
+	}
+
+	links := h.store.GetLinks()
+	found := false
+	for i := 0; i < len(links); i++ {
+		if links[i].Name == autolinkName || links[i].Pattern == autolinkName {
```
mgdelacroix, line 140: I wonder if we should only match by `link.Name` to simplify both the API and the endpoint code, as `link.Name` should be unique and there are no chances of the same parameter matching more than one link

server/api/api.go
```diff
@@ -114,12 +118,85 @@ func (h *Handler) setLink(w http.ResponseWriter, r *http.Request) {
...
+	autolinkName := r.URL.Query().Get(autolinkclient.AutolinkNameQueryParam)
+	if autolinkName == "" {
+		h.handleSendingJSONContent(w, links)
+		return
+	}
+
+	found := false
+	var autolinks []autolink.Autolink
+	for _, link := range links {
+		if link.Name == autolinkName || link.Pattern == autolinkName {
```
mgdelacroix, line 170: Same suggestion here, we should only check for `link.Name`

server/autolinkclient/client.go
```diff
@@ -64,3 +82,47 @@ func (c *Client) Add(links ...autolink.Autolink) error {
...
+		return nil, err
+	}
+
+	req.URL.RawQuery = queryParams.Encode()
+
+	resp, err := c.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
```
mgdelacroix, line 125: This would forbid the caller from reading the response body, so it should be the caller's responsibility to call the close method

server/autolinkclient/client.go
```diff
@@ -64,3 +82,47 @@ func (c *Client) Add(links ...autolink.Autolink) error {
...
+
+	resp, err := c.call("/"+autolinkPluginID+"/api/v1/link", http.MethodGet, nil, queryParams)
+	if err != nil {
+		return nil, err
+	}
+
+	respBody, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
```
mgdelacroix, line 99: This should run after the `resp.StatusCode` check

server/autolinkclient/client.go
```diff
@@ -64,3 +82,47 @@ func (c *Client) Add(links ...autolink.Autolink) error {
...
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("unable to install autolink. Error: %v, %v", resp.StatusCode, string(respBody))
```
mgdelacroix, line 102:
```suggestion
		return nil, fmt.Errorf("unable to get link %s. Error: %v, %v", autolinkName, resp.StatusCode, string(respBody))
```

server/autolinkclient/client.go
```diff
@@ -45,16 +49,30 @@ func (c *Client) Add(links ...autolink.Autolink) error {
...
+	return nil
+}
+
+func (c *Client) Delete(links ...string) error {
+	for _, link := range links {
+		queryParams := url.Values{
+			AutolinkNameQueryParam: {link},
+		}
+
+		resp, err := c.call("/"+autolinkPluginID+"/api/v1/link", http.MethodDelete, nil, queryParams)
```
mgdelacroix, line 72: This comment belongs to the error line below, when checking for status code. The error message should be something like `unable to delete link` instead of `unable to install autolink`

mgdelacroix (changes requested):

server/api/api.go
Kshitij-Katiyar: sure @mgdelacroix Doing this change.

Kshitij-Katiyar:
@mgdelacroix Please re-review

server/autolinkclient/client.go
```diff
@@ -45,22 +49,79 @@ func (c *Client) Add(links ...autolink.Autolink) error {
 			return err
 		}
 
-		req, err := http.NewRequest("POST", "/"+autolinkPluginID+"/api/v1/link", bytes.NewReader(linkBytes))
+		resp, err := c.call("/"+autolinkPluginID+"/api/v1/link", http.MethodPost, linkBytes, nil)
```
mgdelacroix, line 52: After the `c.call` we need to close the `resp`, as it was moved out of the inner function:

```go
resp, err := c.call("/"+autolinkPluginID+"/api/v1/link", http.MethodPost, linkBytes, nil)
if err != nil {
	return err
}
defer resp.Body.Close()
```

server/autolinkclient/client.go
```diff
@@ -45,22 +49,79 @@ func (c *Client) Add(links ...autolink.Autolink) error {
...
+	for _, link := range links {
+		queryParams := url.Values{
+			AutolinkNameQueryParam: {link},
+		}
+
+		resp, err := c.call("/"+autolinkPluginID+"/api/v1/link", http.MethodDelete, nil, queryParams)
 		if err != nil {
 			return err
 		}
-		defer resp.Body.Close()
```
mgdelacroix, line 57: We need to keep this closing of the body

server/api/api.go
```diff
@@ -114,12 +118,85 @@ func (h *Handler) setLink(w http.ResponseWriter, r *http.Request) {
...
+		h.handleError(w, errors.New("autolink name or pattern should not be empty"))
+		return
+	}
+
+	links := h.store.GetLinks()
+	found := false
+	for i := 0; i < len(links); i++ {
+		if links[i].Name == autolinkName {
+			links = append(links[:i], links[i+1:]...)
+			found = true
```
mgdelacroix, line 142: As there is only one possible match here, we can break on found:

```suggestion
			found = true
			break
```

server/autolinkclient/client.go
```diff
@@ -45,22 +49,79 @@ func (c *Client) Add(links ...autolink.Autolink) error {
...
+	}
+
+	var respBody []byte
+	if resp.StatusCode != http.StatusOK {
+		respBody, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return nil, err
+		}
+		return nil, fmt.Errorf("unable to get the link %s. Error: %v, %v", autolinkName, resp.StatusCode, string(respBody))
+	}
```
mgdelacroix, line 103: Apologies for my previous comment, I didn't realise that we were using `respBody` in both success and error cases. It's fine to keep reading the `resp` before performing the status code check:

```suggestion
	respBody, err := ioutil.ReadAll(resp.Body)
	if err != nil {
		return nil, err
	}

	if resp.StatusCode != http.StatusOK {
		return nil, fmt.Errorf("unable to get the link %s. Error: %v, %v", autolinkName, resp.StatusCode, string(respBody))
	}

```

server/api/api.go
```diff
@@ -114,12 +118,85 @@ func (h *Handler) setLink(w http.ResponseWriter, r *http.Request) {
...
 		status = http.StatusOK
 	}
 
+	ReturnStatusOK(status, w)
+}
+
+func (h *Handler) deleteLink(w http.ResponseWriter, r *http.Request) {
+	autolinkName := r.URL.Query().Get(autolinkclient.AutolinkNameQueryParam)
+	if autolinkName == "" {
+		h.handleError(w, errors.New("autolink name or pattern should not be empty"))
```
mgdelacroix, line 133:
```suggestion
		h.handleError(w, errors.New("autolink name should not be empty"))
```

server/api/api.go
```diff
@@ -114,12 +118,85 @@ func (h *Handler) setLink(w http.ResponseWriter, r *http.Request) {
...
+		if link.Name == autolinkName {
+			autolinks = append(autolinks, link)
+			found = true
+		}
+	}
+
+	if !found {
+		h.handleError(w, errors.Errorf("no autolink found with name %s", autolinkName))
+		return
+	}
```
mgdelacroix, line 179: If we're only matching by name and the name is unique, we're only going to find one link per query, right? In that case, this could be something like:

```suggestion
	var autolink *autolink.Autolink = nil
	for _, link := range links {
		if link.Name == autolinkName {
			autolink = link
			break
		}
	}

	if autolink == nil {
		h.handleError(w, errors.Errorf("no autolink found with name %s", autolinkName))
		return
	}
```

And then we should change the client to match the single response, see comments there.

server/autolinkclient/client.go
```diff
@@ -45,22 +49,79 @@ func (c *Client) Add(links ...autolink.Autolink) error {
...
 			respBody, _ := ioutil.ReadAll(resp.Body)
-			return fmt.Errorf("unable to install autolink. Error: %v, %v", resp.StatusCode, string(respBody))
+			return fmt.Errorf("unable to delete the link %s. Error: %v, %v", link, resp.StatusCode, string(respBody))
 		}
 	}
 
 	return nil
 }
+
+func (c *Client) Get(autolinkName string) ([]*autolink.Autolink, error) {
```
mgdelacroix, line 86:
```suggestion
func (c *Client) Get(autolinkName string) (*autolink.Autolink, error) {
```

server/autolinkclient/client.go
```diff
@@ -45,22 +49,79 @@ func (c *Client) Add(links ...autolink.Autolink) error {
...
+	var respBody []byte
+	if resp.StatusCode != http.StatusOK {
+		respBody, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return nil, err
+		}
+		return nil, fmt.Errorf("unable to get the link %s. Error: %v, %v", autolinkName, resp.StatusCode, string(respBody))
+	}
+
+	var response []*autolink.Autolink
```
mgdelacroix, line 105:
```suggestion
	var response *autolink.Autolink
```

mgdelacroix (commented):
Thanks for applying the changes @Kshitij-Katiyar! Some more comments left

mgdelacroix:
@Kshitij-Katiyar could you please add some tests on the endpoints?

Kshitij-Katiyar:
> @Kshitij-Katiyar could you please add some tests on the endpoints?

@mgdelacroix Added unit test cases.

hanzei:
@mgdelacroix @calebroseland Gentle reminder to review this PR. It's needed for the JIRA plugin.

raghavaggarwal2308:
@mgdelacroix Gentle reminder for review

raghavaggarwal2308:
@mgdelacroix Gentke reminder for review

